PR #1, opened 2021-11-09
WASM

mischnic:
Looks like the perf difference got larger since last time I tested.
```
$ node bench.js ../bootstrap.css
cssnano: 1.128s
162289 bytes

esbuild: 33.304ms
163165 bytes

parcel-css: 14.803ms
161398 bytes
```
```
$ CSS_TRANSFORMER_WASM=1 node bench.js ../bootstrap.css
cssnano: 1.350s
162289 bytes

esbuild: 38.626ms
163165 bytes

parcel-css: 29.698ms
161395 bytes
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
-node_modules
-target
 *.node
+node_modules/
+target/
+pkg/
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "bracketSpacing": false,
+  "endOfLine": "lf",
+  "singleQuote": true,
+  "trailingComma": "all"
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bumpalo"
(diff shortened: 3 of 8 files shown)


devongovett:
Maybe because of jemalloc?

Brooooooklyn:
because opt level changed to `s`. And I recommend https://github.com/purpleprotocol/mimalloc_rust which is faster than jemalloc

devongovett:
Ah, yes. Removing `opt-level=s` does improve perf, at least for the native version. For WASM, the performance is roughly the same. It also doesn't seem to affect size for wasm, so I removed it.

Also btw, I had to add this to Cargo.toml to get it to build on M1:

```
[package.metadata.wasm-pack.profile.release]
wasm-opt = false
```

I think wasm-opt isn't built for ARM yet...

devongovett:
> And I recommend https://github.com/purpleprotocol/mimalloc_rust which is faster than jemalloc

I just tried, and at least for me on an M1 Mac, jemalloc is a little faster. I think in Parcel we use mimalloc on Windows though.

mischnic:
Another thing to note is that the first (couple?) invocation(s) of the wasm version are slower because it's not JITed yet but interpreted. So for accurate timings, you need to do something like
```js
  for(let i = 0; i < 10; i++){
    css.transform(opts);
  }
  console.time('parcel-css');
  let res = css.transform(opts);
  console.timeEnd('parcel-css');
  console.log(res.length + ' bytes');
```

PR #11, opened 2021-12-04
Add csso to benchmark

mischnic:
I also tried to find the css files you used for the benchmark in the readme, but the ones in the npm packages (I listed them in the comment) don't seem to be it?

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -2,6 +2,11 @@ const css = require('./native');
 const cssnano = require('cssnano');
 const postcss = require('postcss');
 const esbuild = require('esbuild');
+const csso = require('csso');
+
+// node_modules/animate.css/animate.css
+// node_modules/bootstrap/dist/css/bootstrap.css
+// node_modules/tailwindcss/dist/tailwind.css
 
 let opts = {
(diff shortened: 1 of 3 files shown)


devongovett:
I was using the files in https://github.com/GoalSmashers/css-minification-benchmark. Not sure if we need to duplicate that whole benchmark here. Hopefully https://github.com/GoalSmashers/css-minification-benchmark/pull/101 will get merged.

devongovett:
Also I'm pretty sure csso does unsafe transforms by default (unless the `restructure` option is turned off). i.e. it merges non-adjacent selectors... 😬

mischnic:
Wow, from their readme 

> `restructure`: Disable or enable a structure optimisations.
> 
> `forceMediaMerge`: ... The optimisation is unsafe in general, but should work fine in most cases. Use it on your own risk.

I got the sense that `restructure` was safe.

onigoetz:
The benchmarks here are now up-to-date : https://goalsmashers.github.io/css-minification-benchmark/

However these don't have specific browser targets, so it's not using advanced optimizations

PR #13, opened 2021-12-10
Feature proposal: save and load playground state as URLHash

Shinyaigeek:
## Motivation 🔥 

Hi team 👋 

`parcel-css` is a great library! I wanted to help something, and I watched this project.

As the [`prettier` playground](https://prettier.io/playground/) and [`swc` playground](https://swc.rs/playground) does, 
How about saving the input state of `parcel-css`'s playground as a URLHash and restoring it on load? This will make it easier to report and reproduce issues.

https://user-images.githubusercontent.com/42742053/145522152-f2d25c0c-87e9-4eca-9592-20ac43712370.mov

This is just a proposal, so if you don’t think it is necessary, please be free to close this PR.

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -1,12 +1,84 @@
-import init, {transform} from '../node/pkg';
+import init, {transform} from '../node/pkg/parcel_css_node.js';
 
 let enc = new TextEncoder();
 let dec = new TextDecoder();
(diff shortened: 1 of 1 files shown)


devongovett (approved):
Awesome, thank you!

PR #30, opened 2022-01-02
Compile logical properties

### LINKED ISSUE #23: Downlevel logical properties
https://caniuse.com/css-logical-props

Only supported very recently. It can be down leveled, but might need to be opt-in because this changes the specificity of rules and requires a `dir` attribute to be placed on an ancestor element (since `:dir(...)` is not yet widely supported (https://caniuse.com/css-dir-pseudo).

```css
.foo {
  margin-inline-start: 10px;
}
```

to

```css
[dir="ltr"] .foo {
  margin-left: 10px;
}

[dir="rtl"] .foo {
  margin-right: 10px;
}
```

See https://github.com/csstools/postcss-logical and https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-dir-pseudo-class

devongovett:
Closes #23

Uses CSS variables to compile CSS logical properties for older browsers. This method allows nesting LTR and RTL regions in a page because custom properties are inherited. To work, it requires the `dir` attribute to be set in HTML (since the `:dir` selector is not yet widely supported by browsers).

Input:
```css
.foo {
  border-inline-start: 2px solid red;
  border-inline-end: 5px solid green;
}

.bar {
  border-inline-start: 1px dotted gray;
  border-inline-end: 1px solid black;
}
```

Output:
```css
.foo {
  border-left: var(--border-left-EgL3uq-1);
  border-right: var(--border-right-EgL3uq-1);
}

.bar {
  border-left: var(--border-left-EgL3uq-2);
  border-right: var(--border-right-EgL3uq-2);
}

[dir="ltr"] {
  --border-left-EgL3uq-1: 2px solid red;
  --border-left-EgL3uq-2: 1px dotted gray;
  --border-right-EgL3uq-1: 5px solid green;
  --border-right-EgL3uq-2: 1px solid #000;
}

[dir="rtl"] {
  --border-left-EgL3uq-1: 5px solid green;
  --border-left-EgL3uq-2: 1px solid #000;
  --border-right-EgL3uq-1: 2px solid red;
  --border-right-EgL3uq-2: 1px dotted gray;
}
```


devongovett:
Found a problem with the approach: it doesn't work with variables. It only works if the variables are defined at or above the element with the dir attribute, not if the variable is defined somewhere within.

```css
some-element {
  --radius: 4px;
  border-start-start-radius: var(--radius);
}
```

produces

```css
[dir="ltr"] {
  --top-left-radius: var(--radius);
  --top-right-radius: ;
}

[dir="rtl"] {
  --top-left-radius: ;
  --top-right-radius: var(--radius);
}

some-element {
  --radius: 4px;
  border-top-left-radius: var(--top-left-radius);
  border-top-right-radius: var(--top-right-radius);
}
```

That doesn't work unless `some-element` itself has a `dir` attribute.

## Alternative

```css
[dir="ltr"] {
  --ltr: initial;
  --rtl: ;
}

[dir="rtl"] {
  --rtl: initial;
  --ltr: ;
}

div {
  border-top-left-radius: var(--ltr, var(--radius));
  border-top-right-radius: var(--rtl, var(--radius));
}
```

This takes advantage of the fact that the `initial` keyword in custom properties marks the property as invalid (https://drafts.csswg.org/css-variables/#guaranteed-invalid) which means variable fallback occurs, and whitespace is a valid variable value, in which case it does not.

This also allows the opposite property to be set the same way:

```css
.foo {
  border-start-start-radius: var(--start);
  border-start-end-radius: var(--end);
}
```

produces:

```css
.foo {
  border-top-left-radius: var(--ltr, var(--start)) var(--rtl, var(--end));
  border-top-right-radius: var(--ltr, var(--end)) var(--rtl, var(--start));
}

[dir="ltr"] {
  --ltr: initial;
  --rtl: ;
}

[dir="rtl"] {
  --ltr: ;
  --rtl: initial;
}
```

Both values are applied in a single declaration, but one of them computes to whitespace which is ignored.

PR #53, opened 2022-01-15
feat: added static method StyleSheet::new

### LINKED ISSUE #46: Support StyleSheet::new
I have a use case where I'd like to construct a new `StyleSheet` with an already populated `CssRuleList`. Is this something you would consider the libary to support?

deckchairlabs:
This PR allows Rust users to construct a new `StyleSheet` instance with a static `new` method.

Resolves #46


devongovett (approved):
Thanks!

PR #54, opened 2022-01-15
Simplify comparison compat

Danue1:
In some of the compatibility tables, I found that min and max are the same.
So, I added a part to change from Ord to Eq.


devongovett (approved):
Thanks!

PR #60, opened 2022-01-16
Add support for custom media queries

### LINKED ISSUE #49: Feature request: support custom media queries
The [Media Queries level 5 working draft](https://www.w3.org/TR/mediaqueries-5/#custom-mq) supports "custom media queries". This is one of the features required to use [open-props](https://open-props.style/) in a Rust application, but would be useful overall. Prior art exists in the [PostCSS Custom Media](https://github.com/postcss/postcss-custom-media) plugin. It'd be great if `parcel-css` could support this too. Thanks!

## Example

__input__
```css
/* --modern targets modern devices that support color or hover */
@custom-media --modern (color), (hover);

@media (--modern) and (width > 1024px) {
  .a { color: green; }
}
```

__output__
```css
@media ((color) or (hover)) and (width > 1024px) {
  .a { color: green; }
}
```

## References
- [Open Props](https://open-props.style/)
- [Media Queries level 5 spec](https://www.w3.org/TR/mediaqueries-5/#custom-mq)
- [PostCSS Custom Media plugin](https://github.com/postcss/postcss-custom-media)

devongovett:
Fixes #49.

This adds basic support for the `@custom-media` rule, and referencing custom media queries in `@media` rules. It will error on complex boolean logic with media types (e.g. screen, print, etc.) because that is too hard to compile and seems unlikely to be used. See https://github.com/parcel-bundler/parcel-css/issues/49#issuecomment-1013897432. It will also error on circular media queries, and undefined references.

To do

- [x] Only compile when not supported by targets (like nesting)
- [x] Allow custom media queries to be defined after they are referenced (https://github.com/w3c/csswg-drafts/issues/6956)
- [ ] Figure out how to handle custom media rules defined in imported files?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -380,6 +380,7 @@ dependencies = [
  "lazy_static",
  "parcel_selectors",
  "parcel_sourcemap",
+ "retain_mut",
  "serde",
  "smallvec",
 ]
@@ -616,6 +617,12 @@ checksum = "16abd0c1b639e9eb4d7c50c0b8100b0d0f849be2349829c740fe8e6eb4816429"
 dependencies = [
  "rand_core",
 ]
+
+[[package]]
+name = "retain_mut"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "11000e6ba5020e53e7cc26f73b91ae7d5496b4977851479edb66b694c0675c21"
 
 [[package]]
 name = "rkyv"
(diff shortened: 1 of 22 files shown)


devongovett:
Going to merge this. For now, `@custom-media` rules must be defined in the same file. Supporting `@import` is complex because reading from the file system is not always possible (e.g. WASM). Will think more about that later.

yoshuawuyts:
Fantastic, thank you so much!

Issue #52, opened 2022-01-15
Additional CLI features


devongovett:
As of #44 we have a basic CLI. Currently it can minify and output to stdout or a file, but might be nice to add a few more features.

* `--targets` option. Could accept a browserslist and convert them to Parcel CSS targets with `browserslist-rs` (see #37).
* `--sourcemap` option. Should output a source map file in addition to a css file, and add a `sourceMappingUrl` comment to the CSS.
* `--nesting` option to enable parsing CSS nesting.
* `--css-modules` option to enable CSS modules. Output a JSON file in addition to compiled CSS?
* Ability to compile multiple files at once, either by passing file names or directories.

Shinyaigeek:
I think it is a good idea to add CLI flags to specify `parcel-css`’s feature.

I think it would be also good to add loading a config file `.parcelcssrc` like `.swcrc` to specify `parcel-css`’s feature as well as CLI flag, because in real development, devs may want to specify which ‘parcel-css`’s feature is ON with a static config file. What do you think about this idea?

devongovett:
Maybe, but I feel like at that point you should probably be using another build tool like a bundler. For example, like Parcel [supports](https://github.com/parcel-bundler/parcel-css#with-parcel). I was thinking that the Parcel CSS CLI would mainly be used for very simple use cases, as well as debugging/testing. Will have to think about it more, but I don't want to jump straight to config files if we can help it. 😄

Shinyaigeek:
Well, in fact,  few devs touch `.swcrc` directory and many devs handle swc feature via bundler. It would be better to implement only CLI flag and collect feedback from developers, and if there are many requests for such a config file, we would rethink 😄 .

devongovett:
Maybe we could also use miette to get pretty error messages: https://github.com/zkat/miette

devongovett:
Also need to figure out a nice way of distributing the CLI. Right now you can install with `cargo` but that means you need to compile it yourself. We could potentially publish binaries to npm, or via direct download or something...

numToStr:
Hey, I was wondering are you planning to add subcommands? Or just bunch of flags and options to control the behaviour.

devongovett:
We could add some, but not sure what they would be. Did you have one in mind? Right now there's `parcel_css minify` but maybe it should be `--minify` so you could choose whether to enable it or not.

joelmoss:
[...]

PR #68, opened 2022-01-20
Add sourcemap, nesting, and css-modules CLI flags

rrcobb:
Address parts of #52, but not `--targets` or transforming multiple files.

* makes `--minify` a boolean option, instead of a subcommand
* instead of  `--input-file` option as part of the minify subcommand, input file is now a required positional arg
* moves `--output-file` from subcommand to the top level
* adds `--nesting` boolean option, to enable nesting
* adds `--css-modules` and `--css-modules-output-file` options
* adds `--sourcemap` option
* adds integration tests in rust for the CLI in `tests/cli_integration_tests.rs`

CSS Modules and sourcemap options make what seem to me like decent default choices for where to send their output. 

* Sourcemap requires an output file, since it doesn't make sense to have a sourcemap without a file...
* if there's an output file `out.css`, then the map is written to `out.css.map`
* css modules allows specifying a path for the json output, or writes to `out.json` if there's an output file named `out.css`

I'd be glad for any feedback, and happy to carve it into smaller chunks, in case this PR is too large.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,9 +13,12 @@ struct CliArgs {
   /// The destination file for the output
   #[clap(short, long)]
   output_file: Option<String>,
-  /// Enable CSS nesting in output
+  /// Enable nesting in output
   #[clap(short, long)]
   nesting: bool,
+  /// Enable CSS modules in output
+  #[clap(short, long)]
+  css_modules: bool,
 }
 
 pub fn main() -> Result<(), std::io::Error> {
@@ -24,6 +27,7 @@ pub fn main() -> Result<(), std::io::Error> {
 
   let mut stylesheet = StyleSheet::parse(cli_args.input_file.to_string(), &source, ParserOptions {
     nesting: cli_args.nesting,
+    css_modules: cli_args.css_modules,
     ..ParserOptions::default()
   }).unwrap();
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,20 +5,21 @@ use parcel_css::stylesheet::{MinifyOptions, ParserOptions, PrinterOptions, Style
 #[derive(Parser, Debug)]
 #[clap(author, about, long_about = None)]
 struct CliArgs {
-  /// The CSS file to minify
+  /// Target CSS file
   input_file: String,
+  /// Destination file for the output
+  #[clap(short, long)]
+  output_file: Option<String>,
   /// Minify the output 
   #[clap(short, long)]
   minify: bool,
-  /// The destination file for the output
-  #[clap(short, long)]
-  output_file: Option<String>,
-  /// Enable nesting in output
+  /// Enable parsing CSS nesting
   #[clap(short, long)]
   nesting: bool,
   /// Enable CSS modules in output
   #[clap(short, long, group = "css_modules")]
   css_modules: bool,
+  /// Default: <output_file>.json
   #[clap(long, requires = "css_modules")]
   css_modules_output_file: Option<String>,
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,22 +3,29 @@ use clap::Parser;
 use parcel_css::stylesheet::{MinifyOptions, ParserOptions, PrinterOptions, StyleSheet};
 
 #[derive(Parser, Debug)]
-#[clap(author, version, about, long_about = None)]
+#[clap(author, about, long_about = None)]
 struct CliArgs {
   /// The CSS file to minify
   input_file: String,
   /// Minify the output 
   #[clap(short, long)]
   minify: bool,
-    /// The destination file for the output
+  /// The destination file for the output
   #[clap(short, long)]
   output_file: Option<String>,
+  /// Enable CSS nesting in output
+  #[clap(short, long)]
+  nesting: bool,
 }
 
 pub fn main() -> Result<(), std::io::Error> {
   let cli_args = CliArgs::parse();
   let source = fs::read_to_string(&cli_args.input_file)?;
-  let mut stylesheet = StyleSheet::parse(cli_args.input_file.to_string(), &source, ParserOptions::default()).unwrap();
+
+  let mut stylesheet = StyleSheet::parse(cli_args.input_file.to_string(), &source, ParserOptions {
+    nesting: cli_args.nesting,
+    ..ParserOptions::default()
+  }).unwrap();
 
   if cli_args.minify {
     stylesheet.minify(MinifyOptions::default()).unwrap();
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use std::fs;
+use std::{fs, path, io, ffi};
 use clap::Parser;
 use parcel_css::stylesheet::{MinifyOptions, ParserOptions, PrinterOptions, StyleSheet};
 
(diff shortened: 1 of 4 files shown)


rrcobb:
(rebasing on latest to resolve conflicts)

src/main.rs
```diff
@@ -1,42 +1,137 @@
...
+            }
+        }
 
-  Ok(())
+
+    } else {
+        println!("{}", res.code);
+        if cli_args.css_modules {
+            let css_modules_json = serde_json::to_string(&res.exports)?;
+            println!("{}", css_modules_json);
```
rrcobb, line 118: printing css modules to stdout feels a little weird, but I guess if you add the `--css-modules` flag, you're expecting css modules, so better to get it than not?

Maybe better to require the `--output-file` flag in order to turn on `--css-modules`. Thoughts?

src/main.rs
```diff
@@ -1,42 +1,137 @@
...
+        let path = path::Path::new(&file);
+        if path.extension() == Some(ffi::OsStr::new("json")) {
+            Err(io::Error::new(io::ErrorKind::Other, "Cannot infer a css modules json filename, since the output file extension is '.json'"))
+        } else {
+            // unwrap: the filename option is a String from clap, so is valid utf-8
+            Ok(path.with_extension("json").to_str().unwrap().into())
+        }
+    } else {
+        Ok("styles.json".into())
+    }
```
rrcobb, line 136: on rereading this, I don't actually think we ever infer the filename if we don't have an output file specified; this should take a string and shouldn't have the outer condition

tests/cli_integration_tests.rs
```diff
@@ -0,0 +1,329 @@
...
+use std::process::Command;
+
+fn test_file() -> Result<assert_fs::NamedTempFile, FixtureError> {
+    let file = assert_fs::NamedTempFile::new("test.css")?;
+    file.write_str(
+        r#"
+      .foo {
+        border: none;
+      }
+    "#,
```
rrcobb, line 17: Borrowed test values from `src/lib.rs`

src/main.rs
devongovett: Yeah I think requiring output file for css modules makes sense. Same with source maps, unless we inline the whole thing. I guess the main usecase for stdout is to pipe it somewhere, so probably most useful if that only includes the css.

devongovett:
Skimmed it and it looks great. Thanks for writing tests as well! Will look in more detail tomorrow.

joelmoss:
Does this support sending the output of CSS Modules JSON and CSS to stdout?

devongovett:
How would that work and what is the usecase for that? How would you separate them? I assume the usecase for stdout is piping to another program, but if we mix CSS and JSON that will be harder...

joelmoss:
Yeah so that is the tricky part.

My use case is to call the CLI and pipe it to my app. I'm building a Rails middleware that will intercept requests for CSS files and return the CSS file - basically a proxy. CSS files will be built on demand using the parcel-css CLI.

But (at least in development) it is more efficient and faster to return the contents directly from the app, as opposed to the CLI writing to a file(s) and my app opening and rendering file contents of the file.

Instead, my app calls the CLI and returns exactly what the CLI returned, and/or be able to work with the returned data.

So when this is just CSS, all is well. But when we have JSON and CSS - like we do with CSS modules - then it's tricker. I would still prefer to read/write files.

So maybe in this case, we return everything as json:

```json
{
  "css": ".header_abc123 { color: red; }",
  "modules": {
    "header": "header_abc123"
  }
}
```

This makes it trivial for the caller to work with the returned data.

Hope that helps shed some light on what I'm after. thx

src/main.rs
```diff
@@ -1,42 +1,137 @@
...
-      minify_css(cli_args)
+    let cli_args = CliArgs::parse();
+    let source = fs::read_to_string(&cli_args.input_file)?;
+
+    let filename = path::Path::new(&cli_args.input_file)
+        .file_name()
+        .unwrap()
+        .to_str()
+        .unwrap()
+        .into();
```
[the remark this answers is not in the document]
devongovett: I think we should probably avoid this. For example, if you had two different files in your project each named `index.css` or something but in different folders, the class names could clash. Using the full absolute path (or a relative one from some project root) avoids this.

src/main.rs
```diff
@@ -1,42 +1,137 @@
...
-  output_file: Option<String>,
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct SourceMapJson<'a> {
+  version: u8,
+  mappings: String,
+  sources: &'a Vec<String>,
+  sources_content: &'a Vec<String>,
+  names: &'a Vec<String>
 }
```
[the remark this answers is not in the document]
devongovett: Yeah Think putting it in `parcel_sourcemap` would be good eventually. Doesn't have to happen for this PR though.

devongovett:
Going to merge this and do a few fixups separately. Thanks again so much for the contribution!

src/main.rs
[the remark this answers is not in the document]
rrcobb: Oh, that's a good point, that does seem like it will definitely happen, especially once there's multiple file support!

I guess a pertinent question is whether invoking from different locations, the output should be deterministic. I still think yes (though, it's clear that `file_name` doesn't do the job here). Is there something else to hash that makes sense? The source file contents?

src/main.rs
rrcobb: I'll make this switch! Drop the complexity / printing here, and make an output file required for using the css modules flag.

rrcobb:
> How would that work and what is the usecase for that? How would you separate them? I assume the usecase for stdout is piping to another program, but if we mix CSS and JSON that will be harder...

Yeah, I think piping the css modules json file to stdout is wrong; it's probably better to leave it out unless the output file is specified.

rrcobb:
Oh, @devongovett you already shipped the fixups! Nice! 

lmk if there's other pieces I can chip in on, I'm starting to get a sense for the codebase, and I like it!

devongovett:
Thanks! A few things we could add if you're up for any of them:

* Support for compiling multiple files at once
* Support for `--targets`
* Nicer error messages. Right now we just crash due to the `unwrap` but would be nice to have some pretty code frames e.g. https://github.com/zkat/miette
* Maybe a `--json` option to output CSS + source map + css modules as a JSON object to stdout? I think that might satisfy what @joelmoss is looking for.

I'm also currently working on a very basic `--bundle` option that will inline `@import` rules.

joelmoss:
> Maybe a --json option to output CSS + source map + css modules as a JSON object to stdout? I think that might satisfy what @joelmoss is looking for.

That would awesome 👏

joelmoss:
Just thinking out loud here, but I have an alternative to outputting the CSS and JSON...

I currently use Webpack and mini-css-extract-plugin, which supports CSS modules. It also allows me to configure the format of the class names to module names using the `modules.localIdentName` config option. And this actually has a nice side effect of allowing me to know the module names outside of webpack without needing to compile.

So as an example...

I set the `modules.localIdentName` to `'cssm_[md5:hash:hex:8]'`. Then from this output:

```css
.classOne {
  color: red;
}
```

I get output like this:

```css
.cssm_abcd1234 {
  color: red;
}
```

All as you would expect. But of course, in order to actually find out what the resulting module names are, I can run webpack and get the JSON output of class names to module names.

I have a better way :)

Because I know how the module names are created, i can actually guess what they are without needing to run webpack.

So in Ruby (my language of choice), I can get the module name:

```ruby
css_path = 'path/to/my/css_modules.css'
class_name = 'classOne'
Digest::MD5.hexdigest("#{css_path}\x00#{class_name}")[0, 8]
```
I now know the module name without needing to compile webpack and reading the JSON.

Sorry for the slightly long winded description, but what I'm getting at, is could we have a deterministic way of doing the same with parcel-css? This way, all I would need from parcel-css is the CSS output.

I certainly don't need as many config options as mini-css-extract-plugin has. In fact, I don't need any.

Thoughts?

PR #75, opened 2022-01-28
Add "cli" Cargo feature to make binary-only dependencies optional.

joshstoik1:
I think it would be beneficial for library use cases to make the binary-only dependencies (i.e. `clap`) optional. This PR adds a (default) "cli" feature to that end.

Thanks for the excellent project!


devongovett:
I think this is a good idea. We could probably turn off the CLI feature by default also. I haven't yet decided how best to distribute it, but it will probably be as a binary. Therefore, the most likely case for installing from Cargo will be without the CLI (ie as a library).

PR #91, opened 2022-02-17
Node: Add type inference for TransformResult

ArnaudBarre:
This PR improves nullability of transform and bundle outputs.

```ts
let {code, map} = css.transform({
  filename: 'style.css',
  code: Buffer.from('.foo { color: red }'),
  sourceMap: true,
});
console.log(map.toString()); // Valid in TS without assertion 🎉
```

Note: Prettier seems to not be applied on this file, so I tried to follow formatting conventions manually.

### DIFF
--- a/node/index.d.ts
+++ b/node/index.d.ts
@@ -52,16 +52,26 @@ export interface PseudoClasses {
   focusWithin?: string
 }
 
-export interface TransformResult {
+export interface TransformResult<Options extends BundleOptions = BundleOptions> {
   /** The transformed code. */
   code: Buffer,
   /** The generated source map, if enabled. */
-  map: Buffer | void,
+  map: ConditionalOutput<Options, "sourceMap", Buffer>,
   /** CSS module exports, if enabled. */
-  exports: CSSModuleExports | void,
+  exports: ConditionalOutput<Options, "cssModules", CSSModuleExports>,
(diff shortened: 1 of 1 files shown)


devongovett:
Could you explain a little more about what this does? I've used TS but this is next level stuff. 😅

ArnaudBarre:
I changed to use function overloads. The first solution was more scalable & able to narrow down for false cases, but TS became too lax on parameters and not warnings on bad parameter.

I added tests via tsd (now recommended by MS) that will explain the goal.

Tell me if you want me to do others methods & add more tests or remove the setup for tsd (for information, ~50M node_modules, ~3s run on Mac without m1)

ArnaudBarre:
I quickly tested the script and the generics syntax seems to create syntax errors.
Using this [tool](https://transform.tools/typescript-to-flow), we would need to change this two lines into:
```
declare type WithFlag<Options, Key: $Keys<Options>> = {
  ...Options,
  ...$ObjMapi<{ [k: Key]: any }, <key>(key) => true>,
};
declare type NonNullableTransformResult<Key: $Keys<TransformResult>> = {
  ...TransformResult,
  ...$ObjMapi<
    { [k: Key]: any },
    <key>(key) => Exclude<$ElementType<TransformResult, key>, void>
    >,
};
```
But I can't say if it's valid, I don't know flow at all.

Issue #120, opened 2022-03-22
Correct handling of `<family-name>` in font-family


yisibl:
## Input

```css
.foo {
  font-family: 'sans-serif', 'default';
}
```

## Output

```css
.foo {
  font-family: sans-serif, default;
}
```

![image](https://user-images.githubusercontent.com/2784308/159428125-c04f92f4-27e3-4298-967e-b4676e9ff11c.png)

## Expected

```css
.foo {
  font-family: 'sans-serif', 'default';
}
```

`font-family: sans-serif` means that a generic `sans-serif` font family will be used, while `font-family: 'sans-serif'` (with quotes) refers to an actual font that goes by the name of `sans-serif`.

## [CSS Spec](https://www.w3.org/TR/css-fonts-4/#example-c2604b3d)
> Note: this means that if you really have a font whose name is the same as one of the [<generic-family>](https://www.w3.org/TR/css-fonts-4/#generic-family-value)[](https://www.w3.org/TR/css-fonts-4/#example-c2604b3d) names, or the common-keywords|CSS-wide keyword values, it must be quoted.

> To illustrate this, the following unusual font family names are valid because they are quoted:
font-family: "sans-serif", sans-serif;
font-family: "default", sans-serif;
font-family: "initial", sans-serif;
font-family: "inherit", sans-serif;

The full CSS-wide keyword(common-keywords) and `<generic-family>` in the current specification include:

### CSS-wide keyword

https://www.w3.org/TR/css-values-4/#common-keywords

```
"initial"
"inherit"
"unset"
```

### `<generic-family>`

https://www.w3.org/TR/css-fonts-4/#generic-font-families

```
"serif"
"sans-serif"
"cursive"
"fantasy"
"monospace"
"system-ui"
"emoji"
"math"
"fangsong"
"ui-serif"
"ui-sans-serif"
"ui-monospace"
"ui-rounded"
```

Here's a lint tool: https://mothereff.in/font-family

yisibl:
There are more case test failures:

https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#valid_family_names

## Input
```css
.foo {
  font-family: "Goudy Bookletter 1911", sans-serif;
  font-family: "Red/Black", sans-serif;
  font-family: "test@foo", sans-serif;
  font-family: "#POUND", sans-serif;
  font-family: "Hawaii 5-0", sans-serif;  
}
```

## Output

```css
[...]

PR #121, opened 2022-03-24
fix: `font-family: revert/revert-layer` cannot remove quotation marks

yisibl:
CSS wide keywords are confusingly defined(see https://github.com/w3c/csswg-drafts/issues/7170), we left out `revert` and `revert-layer`, and we can't remove the quotes when these keywords are present.

Improvement #120

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3636,7 +3636,17 @@ mod tests {
     minify_test(".foo { font-family: inherit; }", ".foo{font-family:inherit}");
     minify_test(".foo { font-family: inherit test; }", ".foo{font-family:inherit test}");
     minify_test(".foo { font-family: 'inherit test'; }", ".foo{font-family:inherit test}");
+    minify_test(".foo { font-family: revert; }", ".foo{font-family:revert}");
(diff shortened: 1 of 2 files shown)


yisibl:
`font-family: default` is an invalid value, do we need to do special handling?

image

devongovett:
hmm, the spec says that unquoted family names parse as `<custom-ident>`, which reserves `default`: https://drafts.csswg.org/css-fonts/#family-name-syntax

devongovett (approved):
Thanks!

yisibl:
I've made similar improvements in esbuild https://github.com/evanw/esbuild/pull/2135

yisibl:
> hmm, the spec says that unquoted family names parse as `<custom-ident>`, which reserves `default`: https://drafts.csswg.org/css-fonts/#family-name-syntax

https://www.w3.org/TR/css-values-4/#ref-for-css-wide-keywords

> The CSS-wide keywords are not valid <custom-ident>s. The default keyword is reserved and is also not a valid <custom-ident>

PR #131, opened 2022-04-03
Refactor logical properties to use :dir/:lang

### LINKED ISSUE #19: Downlevel :is
The `:is` selector is fairly new so we should probably downlevel it: https://caniuse.com/css-matches-pseudo. We use it currently to compile nesting as well.

In the general case, this is hard if not impossible to do correctly. It can easily explode in combinations and generate massive CSS, and in some cases might be challenging without knowing the DOM because it's essentially boolean logic (if this selector matches, AND this selector matches). But may work in simple cases. See also https://github.com/postcss/postcss-selector-matches/issues/19#issuecomment-468271010 and https://github.com/csstools/postcss-is-pseudo-class.

Perhaps we could compile to the older `:matches` or `:-webkit-any` syntax at least.

devongovett:
The previous approach to compiling logical properties used CSS variables with some clever tricks to define conditional styles for ltr and rtl. However, it had an issue where CSS variables that are invalid cause the property using them to be reset to their initial value rather than the cascaded value. This is the [invalid at computed value time](https://www.w3.org/TR/css-variables-1/#invalid-at-computed-value-time) part of the spec.

This changes our approach to instead use the `:dir` selector instead. However, this is not supported by any browsers other than Firefox, so we fall back to `:lang` using a hard-coded list of RTL languages. This isn't perfect, but it's possibly the best we can do. Using an attribute selector (e.g. `[dir="ltr"]`) doesn't work right either because it affects nested elements too (e.g. rtl inside ltr inside rtl). `:lang` is inherited correctly, just like `:dir`, and is much more widely supported.

In addition to downleveling `:dir` to `:lang`, we also downlevel `:lang` when it is passed a list of languages rather than a single one. Support for multiple languages was added in Selectors Level 4, and is currently only supported in WebKit. For other browsers, we use `:is` with multiple `:lang` selectors inside, or `:not` for the opposite case.

We also now downlevel `:is` to the older `:-webkit-any` and `:-moz-any` pseudo classes. However, these only support simple selectors (no combinators), so if a combinator is seen, we bail out. Fixes #19.


bakura10:
Hi :),

It seems that when using target>0.5%, the following CSS is generated:

image

text-align: start/end are supported from Safari 3 according to CanIUse, so this should likely not be needed here. Is there a way to keep the simple start/end or at least having a setting here?

bakura10:
Would it be possible to offer a setting to use [dir="rtl"] / [dir="ltr"] instead of those options? I am not sure to understand the issue you're referring to. I have used this for fairly large code and never run into an issue. Nesting a rtl inside an ltr seems like a rather edge case, and lead to much simpler output. As the final compile file code is redistributed, I would love to have a simpler and easier to understand output.

bakura10:
EDIT: apparently I was too broad with >0.5%, and this compiled for browser we do not target. I changed it to "safari 13" and the markup is now much more simple :).

PR #134, opened 2022-04-04
Format rust source code using rustfmt

lucacasonato:
"rustfmt's style is nobody's favourite, but rustfmt is everybody's
favourite."


lucacasonato:
+5k lines 🙈

devongovett (approved):

PR #135, opened 2022-04-04
Add std::error::Error implementation for errors

lucacasonato:
This commit adds `std::error::Error`, and by extension `std::fmt::Display` for
all custom error types. I have depreacted (public) or removed (crate local)
`Error::reason()` functions. Users should use `Error::to_string()` instead or an
alternative formatting routine utilizing `std::fmt::Display` instead.


selectors/parser.rs
```diff
@@ -174,20 +174,14 @@ pub enum SelectorParseErrorKind<'i> {
   NoQualifiedNameInAttributeSelector(Token<'i>),
   EmptySelector,
   DanglingCombinator,
-  NonCompoundSelector,
-  NonPseudoElementAfterSlotted,
-  InvalidPseudoElementAfterSlotted,
-  InvalidPseudoElementInsideWhere,
```
lucacasonato, line 180: These error kinds were unused.

devongovett (approved):
Thanks, looks good!

Was there a reason you lowercased all the error messages?

lucacasonato:
> Was there a reason you lowercased all the error messages?

Yeah, from my experience error messages are usually all lowercase in Rust. For example error messages from std are all lowercase. It also makes them more readable when they get printed through many wrapper layers (e.g `error: bundle error: parse error: unexpected token # at file://src/foo.css:1:1`).

devongovett:
Hmm, this does not match all of Parcel's error messages though. Might have to figure something out for that. 🤔

lucacasonato:
I'll go uppercase them again then - it's not the end of the world if they look a little out of place for Rust users. JS users are the primary consumer I think.

PR #136, opened 2022-04-04
CI on PRs, fix build warning, fixed rust version

lucacasonato:
This commit updates CI to also run on PRs. Additionally CI now fails if
build warnings are encountered.

A `rust-toolchain.toml` file has been added to the root of the repo that
ensures that all collaboraters use the same up-to-date Rust version.
Cargo picks up this file automatically and uses rustup to install and
use the right version of rustc and cargo for each project.

CI now also adheres to this toolchain file through the use of
`dtolnay/rust-toolchain@stable`.

### DIFF
--- a/src/media_query.rs
+++ b/src/media_query.rs
@@ -808,7 +808,7 @@ fn process_condition<'i>(
     }
     MediaCondition::Operation(conditions, _) => {
       let mut res = Ok(true);
-      conditions.retain_mut(|condition| {
+      RetainMut::retain_mut(conditions, |condition| {
         let r = process_condition(loc, custom_media, media_type, qualifier, condition, seen);
         if let Ok(used) = r {
           used
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,12 +1,21 @@
 name: test
+
 on:
   push:
+    branches: [master]
+  pull_request:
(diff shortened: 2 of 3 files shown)


src/media_query.rs
```diff
@@ -808,7 +808,7 @@ fn process_condition<'i>(
     }
     MediaCondition::Operation(conditions, _) => {
       let mut res = Ok(true);
-      conditions.retain_mut(|condition| {
+      RetainMut::retain_mut(conditions, |condition| {
```
lucacasonato, line 811: This change is because the old style caused a build warning in rustc 1.59.0:

```
warning: an associated function with this name may be added to the standard library in the future
   --> src/media_query.rs:811:18
    |
811 |       conditions.retain_mut(|condition| {
    |                  ^^^^^^^^^^
    |
    = note: `#[warn(unstable_name_collisions)]` on by default
    = warning: once this associated item is added to the standard library, the ambiguity may cause an error or change in behavior!
    = note: for more information, see issue #48919 <https://github.com/rust-lang/rust/issues/48919>
    = help: call with fully qualified syntax `retain_mut(...)` to keep using the current method

warning: `parcel_css` (lib) generated 1 warning
```

devongovett (approved):

PR #145, opened 2022-04-07
feat: select version for playground

ahabhgk:
This PR implement select specific version of @parcel/css-wasm for playground, might be useful for finding output differences between versions.

select version for playground

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -1,3 +1,6 @@
+import * as localWasm from '../node/pkg';
+
+let wasm;
 let enc = new TextEncoder();
 let dec = new TextDecoder();
 let inputs = document.querySelectorAll('input[type=number]');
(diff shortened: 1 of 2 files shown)


playground/playground.js
```diff
@@ -1,9 +1,53 @@
...
+          resolve();
+          cleanup(script);
+        },
+      });
+
+      document.head.appendChild(script);   
+    })
+  }
+  
+  await load(`https://cdn.jsdelivr.net/npm/@parcel/css-wasm@${version.value}/parcel_css_node.js`);
```
devongovett, line 47: I think we could probably use dynamic import for this. Looks like Parcel doesn't like absolute URLs there yet, but you could do something like this:

```js
let wasm = await new Function('version', 'return import(`https://cdn.jsdelivr.net/npm/@parcel/css-wasm@${version}/parcel_css_node.js`)')(version.value);
await wasm.default();
return wasm;
```

And then pass the returned wasm object to the `update` function, rather than having a global variable?

devongovett (commented):
Very nice! Had one suggestion.


Also, should we have a way to still run the local version? Might be nice in development especially so you could test out local changes.

playground/playground.js
ahabhgk: Yes, I was using dynamic import for this, but I didn't find how to dynamic import an external url resource in Parcel, so I use this for a kind of fallback.
Will switch to the `new Function` way, it seems more convenient and readable, but the `update` function is used to handle `oninput` and `onchange` event, passing the wasm object is not handy for that, so I'm leaving a `let` variable to hold the wasm object to keep it simple~

ahabhgk:
That nice to have a local version, added it back.

PR #151, opened 2022-04-09
Overwrite zero-sized buffers in JS

mischnic:
A workaround for https://github.com/parcel-bundler/parcel-css/issues/150

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -12,14 +12,41 @@ if (process.platform === 'linux') {
   parts.push('msvc');
 }
 
+let bindings;
+
 if (process.env.CSS_TRANSFORMER_WASM) {
-  module.exports = require(`../pkg`);
+  bindings = require(`../pkg`);
 } else {
   try {
-    module.exports = require(`@parcel/css-${parts.join('-')}`);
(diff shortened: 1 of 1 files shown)


devongovett:
Is it possible to fix on the Rust side or do we have to add a JS wrapper?

devongovett:
I guess not because of the serde usage. Maybe if we were constructing the buffers manually. Probably fine.

mischnic:
I couldn't find a way to return an empty buffer from Rust. But instead returning an buffer containing a single whitespace would work (if you'd prefer that over the JS version)

devongovett:
I think I found another workaround. We can use `String::with_capacity(1)` rather than `String::new()` to ensure a buffer is always allocated.

However, this didn't immediately work because it seems `napi-rs` actually makes a copy: https://github.com/napi-rs/napi-rs/blob/abf99a52ce8a76a84e66d32f66b518588093e4ec/napi/src/js_values/ser.rs#L39. We can avoid this by not using serde to construct the return value, and instead do it ourselves. Might improve perf a little too!

PR #169, opened 2022-05-01
Add method to lazily compute property source ranges

### LINKED ISSUE #168: How can we get location of each AST node?
From the [Rust doc](https://docs.rs/parcel_css/1.0.0-alpha.24/parcel_css/index.html), it looks that only rules contain a field called `location`, while other AST nodes don't. How can I get location for any type of AST node?

devongovett:
Closes #168

We don't store source locations for every property/value currently, because that would use a lot of additional memory, and it's rare to need source locations except in errors. So this adds a method to lazily compute the source locations of the keys/values of declarations within a style rule. Essentially, it performs a very lightweight additional parsing pass to first find the stored location of the parent style rule, and then finding the declaration by index from there.

This approach is somewhat error prone (e.g. if you modified the style sheet) so I'm not sure if it's a good idea or not. Alternatives would be to store source locations for everything, which I'd prefer not to do if we can help it, or somehow optionally store them externally (but I'm not sure how).

### DIFF
--- a/src/declaration.rs
+++ b/src/declaration.rs
@@ -1,5 +1,7 @@
 //! CSS declarations.
 
+use std::ops::Range;
+
 use crate::context::PropertyHandlerContext;
 use crate::error::{ParserError, PrinterError};
 use crate::parser::ParserOptions;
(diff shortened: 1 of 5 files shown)


g-plane:
> somehow optionally store them externally

Is it possible that providing something like hooks when parsing then let downstream developers to do something else, such as storing source locations?

devongovett:
Yeah possible, but how would you look them up? Making the nodes hashable or something wouldn't be enough because two nodes could be equal in different parts of the tree, but have different locations. We'd need to store at least some identifier in each node I guess.

g-plane:
I see. Is it possible that calculating source locations for specific type of node in downstream code? For example, just for functions or colors.

g-plane (approved):

devongovett:
Going to merge this for now. Will continue to think about ways to get more granular information in the future.

PR #176, opened 2022-05-11
Serde support

### LINKED ISSUE #143: Implement `serde::Serialize` on AST
This would be very useful for the CSSOM implementation in Deno (also see #5). Many `cssparser` structs already have `serde::Serialize` implementations when you enable the `serde` feature for `cssparser`.

If the binary size increase is too much, we can flag it with a crate feature.

I'll try to open a PR later this week.

devongovett:
Closes #143

This adds serde serialization and deserialization support to all AST nodes, behind a feature flag. Appears to work, but need to verify the output.

cc. @lucacasonato @johno

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,4 +18,4 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v1
       - run: cargo fmt
-      - run: cargo test
+      - run: cargo test --all-features
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,6 +273,7 @@ dependencies = [
  "phf",
  "proc-macro2",
  "quote",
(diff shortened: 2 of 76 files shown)


devongovett:
Did some testing and verified the output seems reasonable. There are certainly improvements we can make (e.g. a nicer format for vendor prefixes in properties, more detailed serialization of selector components, etc.), but it seems usable for now.

To use, you need to enable the `serde` feature flag for the crate. You can also test the example in the readme which parses the provided filename and outputs JSON to stdout.

```shell
cargo run --example serialize --features serde test.css | jless
```

The format follows the documented rust structs, but with kebab case for enum variants (matching CSS itself), and a few other minor differences. Not sure if possible, but a future improvement could be to somehow generate TypeScript definitions for the format.

Going to merge as MVP. Please let me know if you have feedback!

Issue #174, opened 2022-05-07
Custom resolver


dgp1130:
Hi, I'm trying to use `@parcel/css` to bundle CSS files, however there doesn't seem to be a way to provide a custom resolver for `@import` statements, the package appears to [directly use the real file system](https://github.com/parcel-bundler/parcel-css/blob/1b947b480674a170fcac22af0a730b70cfce93a0/node/src/lib.rs#L129-L130). This is a challenge for my use case where I'm hoping to use Parcel for CSS bundling in a [Bazel](https://bazel.build/) workspace. Bazel is a build tool which puts generated files in different directories from the source code, meaning that `@import './foo.css'` could resolve to `./foo` or `$project_root/dist/bin/path/to/dir/foo.css`, or even a few other paths. I would like to configure `@parcel/css` to resolve these paths correctly, but AFAICT, this package doesn't have any direct support for that. See https://github.com/dgp1130/rules_prerender/issues/46 and https://github.com/dgp1130/rules_prerender/commit/7e995d6c41a4b4bbf1235903a2fd400a47f710d3 for a little more info about the use case and my attempt to implement it.

It does seem that [custom resolvers are supported in `.parcelrc`](https://parceljs.org/features/plugins/#resolvers), but I don't get the impression `@parcel/css` has any understanding of that configuration file. I'm unclear how the `parcel` package supports custom CSS resolvers if `@parcel/css` doesn't, but maybe they don't compose each other in the way I think they do? If there is a more appropriate package or different level of abstraction I should be using which would support resolvers, please let me know and I'm happy to try that.

If it does make sense to support custom resolvers in `@parcel/css`, I'm happy to put in the effort to contribute it. I *think* we would just need to [update `BundleConfig` to accept a `SourceResolver`](https://github.com/parcel-bundler/parcel-css/blob/1b947b480674a170fcac22af0a730b70cfce93a0/node/src/lib.rs#L127-L130) from JS and then use that instead of the `FileResolver`. All the inputs and outputs would be strings, so serializing into and out-of JS shouldn't be too much of an issue. I think it's possible, though I haven't done too much with Rust or WASM, so any pointers or suggestions here would be appreciated.

Is this a contribution you would be interested in accepting or am I going about this problem in the wrong manner?

devongovett:
[...]

PR #177, opened 2022-05-13
Add `resolve()` method to `SourceProvider`

dgp1130:
This adds a new `resolve()` method to `SourceProvider` which is responsible for resolving an `@import` specifier into a file path. `FileProvider` still uses the default behavior of assuming the specifier is a relative path and joining it with the path of the originating file. This provides a hook for custom `SourceProviders` to implement their own application-specific resolution logic.

This required a couple minor refactors in the `fs!` and `error_test()` utilities to make them more flexible and support testing a custom resolver.

I've done a little Rust, but I'm not terribly familiar with it, so please let me know if I'm doing anything unidiomatic. Some of the areas I struggled with and am not convinced I found the ideal solution:
1. `error_test()` accepts a callback which runs assertions. This was the only way I could reuse the function to assert different error kinds, since the only way to compare the `BundleErrorKind` enum is with `matches!()`, and a match expression can't really work as a function input. I tried to return the error instead, but that ran into ownership issues since it outlived the bundler, and I couldn't find a good solution there.
2. I wanted the bundler to automatically prepend any resolver errors with "Failed to resolve {}:\n...", so that the context of the failure is always present, even if the resolver itself gave a bad error message. Unfortunately I couldn't find a good way of wrapping the error message from the resolver since it might not be an `IOError` with a `NotFound` kind, so it may not have an error message to wrap, and reaching into that felt a little too intrusive.

Refs #174.

### DIFF
--- a/src/bundler.rs
+++ b/src/bundler.rs
@@ -574,12 +574,14 @@ mod tests {
       .code
   }
 
-  fn error_test(fs: TestProvider, entry: &str) {
+  fn error_test(fs: TestProvider, entry: &str, maybe_cb: Option<Box<dyn FnOnce(BundleErrorKind) -> ()>>) {
     let mut bundler = Bundler::new(&fs, None, ParserOptions::default());
     let res = bundler.bundle(Path::new(entry));
     match res {
       Ok(_) => unreachable!(),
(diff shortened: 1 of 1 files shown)


devongovett (approved):
Thanks, this looks great!

PR #182, opened 2022-05-22
Add support for system-ui fallbacks

onigoetz:
The `system-ui` font family is a special font family that will be changed to the platform's default font family.

This feature will add fallback font families when unsupported

```css
.foo {
  font-family: Helvetica, system-ui, sans-serif;
}

.bar {
  font: 100%/1.5 Helvetica, system-ui, sans-serif;
}
```

becomes

```css
.foo {
  font-family: Helvetica, system-ui, AppleSystem, BlinkMacSystemFont, Segoe UI, Roboto, Noto Sans, Ubuntu, Cantarell, Helvetica Neue, sans-serif;
}

.bar {
  font: 100% / 1.5 Helvetica, system-ui, AppleSystem, BlinkMacSystemFont, Segoe UI, Roboto, Noto Sans, Ubuntu, Cantarell, Helvetica Neue, sans-serif;
}
```

The list of fonts comes from this postcss plugin: https://www.npmjs.com/package/postcss-font-family-system-ui

These are my very first lines of Rust, I'm open to feedback if something isn't idiomatic or could be made more efficiently or anything really :)

### DIFF
--- a/scripts/build-prefixes.js
+++ b/scripts/build-prefixes.js
@@ -168,7 +168,8 @@ let cssFeatures = [
   'css-rrggbbaa',
   'css-nesting',
   'css-not-sel-list',
-  'css-has'
+  'css-has',
+  'font-family-system-ui'
 ];
 
 let compat = new Map();
@@ -315,11 +316,12 @@ use serde::{Deserialize, Serialize};
 /// use parcel_css::targets::Browsers;
 ///
 /// let targets = Browsers {
-///   safari: (13 << 16) | (2 << 8),
+///   safari: Some((13 << 16) | (2 << 8)),
 ///   ..Browsers::default()
 /// };
 /// \`\`\`
 #[derive(Serialize, Debug, Deserialize, Clone, Copy, Default)]
+#[allow(missing_docs)]
 pub struct Browsers {
   pub ${allBrowsers.join(': Option<u32>,\n  pub ')}: Option<u32>
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4686,6 +4686,42 @@ mod tests {
       "@font-face { font-family: 'revert-layer'; }",
       "@font-face{font-family:\"revert-layer\"}",
     );
+
+    prefix_test(
+      r#"
+      .foo {
+        font-family: Helvetica, system-ui, sans-serif;
+      }
+    "#,
+      indoc! {r#"
+      .foo {
+        font-family: Helvetica, system-ui, AppleSystem, BlinkMacSystemFont, Segoe UI, Roboto, Noto Sans, Ubuntu, Cantarell, Helvetica Neue, sans-serif;
+      }
+    "#
+      },
+      Browsers {
+        safari: Some(8 << 16),
+        ..Browsers::default()
+      },
+    );
+
+    prefix_test(
+      r#"
+      .foo {
+        font: 100%/1.5 Helvetica, system-ui, sans-serif;
+      }
+    "#,
+      indoc! {r#"
+      .foo {
+        font: 100% / 1.5 Helvetica, system-ui, AppleSystem, BlinkMacSystemFont, Segoe UI, Roboto, Noto Sans, Ubuntu, Cantarell, Helvetica Neue, sans-serif;
+      }
+    "#
+      },
+      Browsers {
+        safari: Some(8 << 16),
+        ..Browsers::default()
+      },
+    );
   }
 
   #[test]
--- a/src/properties/font.rs
+++ b/src/properties/font.rs
@@ -1,6 +1,7 @@
 //! CSS properties related to fonts.
 
 use super::{Property, PropertyId};
+use crate::compat::Feature;
 use crate::context::PropertyHandlerContext;
 use crate::declaration::{DeclarationBlock, DeclarationList};
 use crate::error::{ParserError, PrinterError};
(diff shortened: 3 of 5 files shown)


scripts/build-prefixes.js
```diff
@@ -315,11 +316,12 @@ use serde::{Deserialize, Serialize};
 /// use parcel_css::targets::Browsers;
 ///
 /// let targets = Browsers {
-///   safari: (13 << 16) | (2 << 8),
+///   safari: Some((13 << 16) | (2 << 8)),
 ///   ..Browsers::default()
 /// };
 /// \`\`\`
 #[derive(Serialize, Debug, Deserialize, Clone, Copy, Default)]
+#[allow(missing_docs)]
```
onigoetz, line 324: I changed these lines to be the same as what the output already contains

src/properties/font.rs
```diff
@@ -908,6 +912,42 @@ impl<'i> PropertyHandler<'i> for FontHandler<'i> {
...
+
+    // Insert values right after "system-ui"
+    let mut unwrapped_family = family.unwrap();
+    unwrapped_family.splice(
+      (position + 1)..(position + 1),
+      &mut default_system_fonts.iter().cloned(),
+    );
+
+    return Some(unwrapped_family);
+  }
```
devongovett, line 946: You could avoid the unwraps here, and the duplicate contains/position check with something like this:

```rust
if let Some(families) = &mut family {
  if let Some(position) = families.iter().position(|v| *v == system_ui) {
    families.splice(
      (position + 1)..(position + 1),
      DEFAULT_SYSTEM_FONTS
        .iter()
        .map(|name| FontFamily::FamilyName(CowArcStr::from(*name))),
    );
  }
}
```

You'll need to add `mut` to the definition of `family` in the function arguments for this to work

src/properties/font.rs
```diff
@@ -908,6 +912,42 @@ impl<'i> PropertyHandler<'i> for FontHandler<'i> {
...
 }
 
+/// [`system-ui`](https://www.w3.org/TR/css-fonts-4/#system-ui-def) is a special generic font family
+/// It is platform dependent but if not supported by the target will simply be ignored
+/// This list is an attempt at providing that support
+#[inline]
+fn compatible_font_family(family: Option<Vec<FontFamily>>, is_supported: bool) -> Option<Vec<FontFamily>> {
+  let system_ui = &FontFamily::Generic(GenericFontFamily::SystemUI);
+
+  let default_system_fonts: [FontFamily; 8] = [
```
devongovett, line 922: With the suggestion below, this could be a const of strings instead:

```rust
const DEFAULT_SYSTEM_FONTS: &[&str] = &[
  // #1: Supported as the -apple-system value (only on Mac)
  "AppleSystem",
  // #2: Supported as the 'BlinkMacSystemFont' value (only on Mac)
  "BlinkMacSystemFont",
  "Segoe UI",  // Windows >= Vista
  "Roboto",    // Android >= 4
  "Noto Sans", // Plasma >= 5.5
  "Ubuntu",    // Ubuntu >= 10.10
  "Cantarell", // GNOME >= 3
  "Helvetica Neue",
];
```

src/properties/font.rs
```diff
@@ -908,6 +912,42 @@ impl<'i> PropertyHandler<'i> for FontHandler<'i> {
   }
 }
 
+/// [`system-ui`](https://www.w3.org/TR/css-fonts-4/#system-ui-def) is a special generic font family
+/// It is platform dependent but if not supported by the target will simply be ignored
+/// This list is an attempt at providing that support
+#[inline]
+fn compatible_font_family(family: Option<Vec<FontFamily>>, is_supported: bool) -> Option<Vec<FontFamily>> {
+  let system_ui = &FontFamily::Generic(GenericFontFamily::SystemUI);
```
devongovett, line 920: Could be a const

```rust
const SYSTEM_UI: FontFamily = FontFamily::Generic(GenericFontFamily::SystemUI);
```

src/properties/font.rs
```diff
@@ -908,6 +912,42 @@ impl<'i> PropertyHandler<'i> for FontHandler<'i> {
   }
 }
 
+/// [`system-ui`](https://www.w3.org/TR/css-fonts-4/#system-ui-def) is a special generic font family
+/// It is platform dependent but if not supported by the target will simply be ignored
+/// This list is an attempt at providing that support
+#[inline]
+fn compatible_font_family(family: Option<Vec<FontFamily>>, is_supported: bool) -> Option<Vec<FontFamily>> {
```
devongovett, line 919: You could either return early here if `is_supported` is true, or not even call this function in that case. Either way it would avoid some extra checks.

devongovett (changes requested):
Thanks for the PR! A few comments that might help make it a bit more idiomatic Rust, but what you did was by no means wrong.

Would you also mind adding a test? `lib.rs` has most of them. It's pretty long but you could add a test to the `test_font` function in there.

onigoetz:
Thanks for the feedback, will do that tonight

Indeed I can add tests, forgot to ask you where to put them.

src/properties/font.rs
onigoetz: I chose the early return

src/properties/font.rs
onigoetz: done

src/properties/font.rs
onigoetz: done. I tried to extract it before but by keeping the `FontFamily` and the compiler told me that `CowArcStr` could not be in a static context. looks better indeed

src/properties/font.rs
onigoetz: Truly a fascinating language, thanks, I'll try that construct next time

src/lib.rs
```diff
@@ -4686,6 +4686,42 @@ mod tests {
...
+
+    prefix_test(
+      r#"
+      .foo {
+        font-family: Helvetica, system-ui, sans-serif;
+      }
+    "#,
+      indoc! {r#"
+      .foo {
+        font-family: Helvetica, system-ui, AppleSystem, BlinkMacSystemFont, Segoe UI, Roboto, Noto Sans, Ubuntu, Cantarell, Helvetica Neue, sans-serif;
```
devongovett, line 4698: I guess system-ui should be first?

onigoetz: I put it second on purpose, this makes sure that the fonts are added at the right place and don't remove the existing. Also, not everybody has Helvetica so it could make sense to want Helvetica but fallback to whatever system font the person has.

devongovett: Ah, sorry, I misread the input. Thanks!

devongovett (approved):

PR #204, opened 2022-06-13
Types: Add method overload for transform & bundle functions

ArnaudBarre:
supersede #91 

Here is a simpler version for functions overloads (like [esbuild](https://github.com/evanw/esbuild/blob/master/lib/shared/types.ts#L466-L470)) that seems to be a valid flow syntax (But I'm not sure it would provide as much type information than TS). 

I'm using `| undefined` because it very uncommon to use void in that case and creates false positive with eslint rules that uses type information.


ArnaudBarre:
What do you think about committing the current version of flow types, to allow TS types to evolve without the need for a script to translate them into flow types?

PR #210, opened 2022-06-17
fix: add libc fields only to platforms that have libc

yisibl:
Add the libc field in a clearer and more explicit way.
After this PR, `linux-gnueabihf` also no longer adds the `libc` field.
### Before

```json
{
  "name": "@parcel/css-linux-arm-gnueabihf",
  "libc": [
    "gnueabihf"
  ]
}
```

Related PRs: https://github.com/parcel-bundler/parcel-css/pull/205, https://github.com/parcel-bundler/parcel-css/pull/208

### DIFF
--- a/scripts/build-npm.js
+++ b/scripts/build-npm.js
@@ -2,15 +2,37 @@ const fs = require('fs');
 const pkg = require('../package.json');
 
 const dir = `${__dirname}/..`;
+
+// Add `libc` fields only to platforms that have libc(Standard C library).
 const triples = [
-  'x86_64-apple-darwin',
-  'x86_64-unknown-linux-gnu',
-  'x86_64-pc-windows-msvc',
-  'aarch64-apple-darwin',
-  'aarch64-unknown-linux-gnu',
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
@yisibl the libc field on `linux-arm-gnueabihf` should be `glibc`.

Brooooooklyn:
@yisibl the libc field on `linux-arm-gnueabihf` should be `glibc`.

yisibl:
@Brooooooklyn Thanks! It looks like the [napi-rs template](https://github.com/napi-rs/package-template/blob/0e2b015e97aa8b9aac9cdad5b8e2e375273f0981/npm/linux-arm-gnueabihf/package.json#L4-L9) needs to be updated as well?

Brooooooklyn:
@yisibl I don't think the `libc` field in `linux-arm-gnueabihf` is necessary, since there is only `glibc` runtime for `linux-arm`.

yisibl:
@devongovett PTAL, thanks!

devongovett (approved):

PR #220, opened 2022-07-01
fix: support importing via ES modules

aminya:
This allows using parcel-css from ES modules. Some build tools (e.g. Astro) require ES modules.

### DIFF
--- a/node/index.mjs
+++ b/node/index.mjs
@@ -1,0 +1,32 @@
+let parts = [process.platform, process.arch];
+if (process.platform === 'linux') {
+  const {MUSL, family} = await import('detect-libc');
+  if (family === MUSL) {
+    parts.push('musl');
+  } else if (process.arch === 'arm') {
+    parts.push('gnueabihf');
+  } else {
+    parts.push('gnu');
+  }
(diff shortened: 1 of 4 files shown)


aminya:
`index.mjs` can also be exported from `package.json`

node/index.mjs
```diff
@@ -0,0 +1,32 @@
...
+  const require = createRequire(import.meta.url);
+
+  try {
+    default_export = require(`@parcel/css-${parts.join('-')}`);
+  } catch (err) {
+    default_export = require(`../parcel-css.${parts.join('-')}.node`);
+  }
+}
+
+export default default_export;
```
devongovett, line 29: Maybe some of the above logic could be deduplicated with the `index.js` file? 

```js
import {createRequire} from 'module';
const require = createRequire(import.meta.url);

export default require('./index.js');
```

aminya: What about the reverse? Generating the old commonjs file from the mjs file?

devongovett: I guess it could be, but it would just be CommonJS to load the native module regardless.

PR #223, opened 2022-07-11
test: add more @container tests

yisibl:


src/rules/container.rs
```diff
@@ -76,7 +76,13 @@ impl<'a, 'i> ToCssWithContext<'a, 'i> for ContainerRule<'i> {
       name.to_css(dest)?;
       dest.write_char(' ')?;
     }
+
+    // Don't downlevel range syntax in container queries.
+    let mut targets = None;
+    std::mem::swap(&mut targets, &mut dest.targets);
     self.condition.to_css(dest)?;
+    std::mem::swap(&mut targets, &mut dest.targets);
```
yisibl, line 84: Hi, I'm a little confused, why is it executed twice here?

Issue #254, opened 2022-08-17
Support the new `tech()` and `format()` syntax in `@font-face`


yisibl:
The specification has updated the syntax, which has been implemented in [Firefox 105](https://bugzilla.mozilla.org/show_bug.cgi?id=1715546).

e.g.

```js
// format() function must precede tech() if both are present
{ src: 'url("foo.ttf") format(opentype) tech(feature-opentype)', valid: true },
{ src: 'url("foo.ttf") tech(feature-opentype) format(opentype)', valid: false },
```
Spec: https://drafts.csswg.org/css-fonts/#font-face-src-parsing

```
<url> [ format(<font-format>)]? [ tech( <font-tech>#)]? | local(<font-face-name>)
<font-format>= [<string> | collection | embedded-opentype | opentype
 | svg | truetype | woff | woff2 ]
<font-tech>= [<font-feature-tech> | <color-font-tech>
 | variations | palettes | incremental ]
<font-feature-tech>= [feature-opentype | feature-aat | feature-graphite]
<color-font-tech>= [color-COLRv0 | color-COLRv1 | color-SVG | color-sbix | color-CBDT ]
```

See also: 

- https://github.com/web-platform-tests/wpt/commit/9f8a6ccc41aa725e8f51f4f096f686313bb88d8d
- https://github.com/mozilla/gecko-dev/commit/bc708c85cdeec156346ffbfe1eb665414ea718ef

PR #255, opened 2022-08-17
refactor(rules/font_face): change supports to tech

CGQAQ:
Closes: #254 

https://github.com/parcel-bundler/parcel-css/blob/5a6801ce26f9ae9d07d44d3427de4a6b48663261/src/lib.rs#L9425-L9430
should we have this test, if we should, please tell me how to make it pass correctly, thanks

I tried to propagate the error up
https://github.com/parcel-bundler/parcel-css/blob/5a6801ce26f9ae9d07d44d3427de4a6b48663261/src/rules/font_face.rs#L70-L85

But here the error have been ignored again
https://github.com/parcel-bundler/parcel-css/blob/5a6801ce26f9ae9d07d44d3427de4a6b48663261/src/rules/font_face.rs#L430-L435

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9469,8 +9469,8 @@ mod tests {
     );
     // multiple tech
     minify_test(
-      "@font-face {src: url(\"test.woff\") format(woff) tech(features-opentype, color-sbix);}",
-      "@font-face{src:url(test.woff)format(\"woff\")tech(features-opentype,color-sbix)}",
+      "@font-face {src: url(\"test.woff\") format(woff) tech(feature-opentype, color-sbix);}",
+      "@font-face{src:url(test.woff)format(\"woff\")tech(feature-opentype,color-sbix)}",
     );
     minify_test(
       "@font-face {src: url(\"test.woff\")   format(woff)    tech(incremental, color-svg, feature-graphite, feature-aat);}",
@@ -9478,13 +9478,13 @@ mod tests {
     );
     // format() function must precede tech() if both are present
     minify_test(
-      "@font-face {src: url(\"foo.ttf\") format(opentype) tech(color-COLRv1);}",
-      "@font-face{src:url(foo.ttf)format(\"opentype\")tech(color-COLRv1)}",
+      "@font-face {src: url(\"foo.ttf\") format(opentype) tech(color-colrv1);}",
+      "@font-face{src:url(foo.ttf)format(\"opentype\")tech(color-colrv1)}",
     );
     // only have tech is valid
     minify_test(
       "@font-face {src: url(\"foo.ttf\") tech(color-SVG);}",
-      "@font-face{src:url(foo.ttf)tech(color-SVG)}",
+      "@font-face{src:url(foo.ttf)tech(color-svg)}",
     );
     // CGQAQ: if tech and format both presence, order is matter, tech before format is invalid
     // but now just return raw token, we don't have strict mode yet.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9448,7 +9448,7 @@ mod tests {
       "@font-face{src:url(test.woff)format(\"woff\")}",
     );
     minify_test(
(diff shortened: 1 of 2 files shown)


devongovett:
Thanks, will look closer tomorrow. 

> But here the error have been ignored again

Yeah, if a property fails to parse, it is stored as raw tokens. This way, if a new feature is added that we don't support yet, we don't fail to parse, it'll just potentially be less optimally minified. I think eventually we might have a strict mode to fail in this case, and we might also emit warnings otherwise.

src/lib.rs
```diff
@@ -9387,34 +9387,59 @@ mod tests {
...
     minify_test("@font-face {src: local(Test);}", "@font-face{src:local(Test)}");
     minify_test("@font-face {src: local(Foo Bar);}", "@font-face{src:local(Foo Bar)}");
+
     minify_test(
       "@font-face {src: url(\"test.woff\") format(woff);}",
       "@font-face{src:url(test.woff)format(\"woff\")}",
     );
     minify_test(
-      "@font-face {src: url(\"test.woff\") format(woff), url(test.ttf) format(truetype);}",
+      "@font-face {src: url(\"test.woff\") format(woff), url(test.ttf) format(\"truetype\");}",
```
yisibl, line 9396: Let's not change the original test cases.

src/rules/font_face.rs
```diff
@@ -104,20 +110,46 @@ pub struct UrlSource<'i> {
...
+            kind: ParseErrorKind::Basic(BasicParseErrorKind::AtRuleBodyInvalid),
+            location: input.current_source_location(),
+          });
+        } else if tech_location.line == location.line && tech_location.column < location.column {
+          return Err(ParseError {
+            kind: ParseErrorKind::Basic(BasicParseErrorKind::AtRuleBodyInvalid),
+            location: input.current_source_location(),
+          });
+        }
+      }
```
devongovett, line 145: How could these checks ever happen? `format_location` will always be less than `tech_location` since you already parsed it before if it exists.

src/rules/font_face.rs
```diff
@@ -133,57 +165,18 @@ impl<'i> ToCss for UrlSource<'i> {
...
-    if !self.supports.is_empty() {
-      dest.write_str(" supports ")?;
-      let mut first = true;
-      for technology in &self.supports {
-        if first {
-          first = false;
-        } else {
-          dest.write_char(' ')?;
+    let tech_len = self.tech.len();
+    if tech_len > 0 {
```
devongovett, line 170:
```suggestion
    if !self.tech.is_empty() {
```

src/rules/font_face.rs
```diff
@@ -133,57 +165,18 @@ impl<'i> ToCss for UrlSource<'i> {
...
+      dest.whitespace()?;
+      dest.write_str("tech(")?;
+      for i in 0..tech_len {
+        self.tech[i].to_css(dest)?;
+        if tech_len - 1 != i {
+          dest.write_char(',')?;
         }
-        technology.to_css(dest)?;
       }
```
devongovett, line 178: you can do `self.tech.to_css(dest)?;` here - `Vec` implements the `ToCss` trait if its contents does.

src/rules/font_face.rs
```diff
@@ -257,75 +251,82 @@ impl<'i> ToCss for FontFormat<'i> {
...
 /// [src](https://drafts.csswg.org/css-fonts/#src-desc)
 /// property of an `@font-face` rule.
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(
   feature = "serde",
   derive(serde::Serialize, serde::Deserialize),
-  serde(tag = "type", content = "value", rename_all = "kebab-case")
+  serde(tag = "type", content = "value")
 )]
 pub enum FontTechnology {
```
devongovett, line 263: You could probably use the `enum_property` macro to simplify this if you want. It would implement `Parse` and `ToCss` for you.

PR #256, opened 2022-08-18
C bindings

devongovett:
This adds C bindings for Parcel CSS, so that it can be used from programming languages other than Rust. So far, the basics of parsing, transforming, and printing stylesheets are supported. Eventually we might be able to expose information about rules and such as well if needed.

cc. @Jarred-Sumner


Jarred-Sumner:
wow this is great

PR #259, opened 2022-08-22
Playground: Fix wrong text color in output panel

marvinhagemeister:
Before:
Screenshot 2022-08-22 at 18 35 46

After:
Screenshot 2022-08-22 at 18 35 53


devongovett:
Thanks

PR #260, opened 2022-08-24
fix: Missing semicolon in Custom Media Queries

### LINKED ISSUE #258: Preserve unused `@custom-media`
Hiya!

I'm bundling a CSS file that includes shared custom properties and custom media. Something like this:

```css
/** media-queries.css **/
@custom-media --breakpoint-large (width >= 75em);
```

```css
/** variables.css **/
:root {
--color-blue: blue;
}
```

```css
/** main.css **/
@import './variables.css';
@import './media-queries.css';
```
I bundle all the files using: 
```
parcel-css ./src/main.css --custom-media --bundle --targets 'last 2 versions' -o ./dist/global.css"
```
And this is the output:
```css
/** global.css **/
:root {
--color-blue: blue;
}
```
I'm assuming because the `@custom-media` is not used in the `global.css` it is removed from the file.

Is there a way to preserve the `@custom-media` declaration?

If I don't use the `--targets` flag in the CLI I almost get what I need but unfortunately semicolons are removed from the declaration:

```css
/** global.css **/
:root {
--color-blue: blue;
}
@custom-media --breakpoint-large (width >= 75em)
```

Causing it to not be parsed at all.

Marabyte:
fix #258

Adds a semicolon at the end of the `@custom-media` declaration.

@devongovett thank you for pointing me to the right file!

### DIFF
--- a/tests/cli_integration_tests.rs
+++ b/tests/cli_integration_tests.rs
@@ -366,3 +366,22 @@ fn targets() -> Result<(), Box<dyn std::error::Error>> {
 
   Ok(())
 }
+
+#[test]
+fn preserve_custom_media() -> Result<(), Box<dyn std::error::Error>> {
+  let file = assert_fs::NamedTempFile::new("test.css")?;
+  file.write_str(
+    r#"
+      @custom-media --foo print;
+    "#,
+  )?;
+
+  let mut cmd = Command::cargo_bin("parcel_css")?;
+  cmd.arg(file.path());
+  cmd.arg("--custom-media");
+  cmd.assert().success().stdout(predicate::str::contains(indoc! {r#"
+    @custom-media --foo print;
+  "#}));
+
+  Ok(())
+}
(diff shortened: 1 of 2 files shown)


tests/cli_integration_tests.rs
```diff
@@ -366,3 +366,22 @@ fn targets() -> Result<(), Box<dyn std::error::Error>> {
...
+    "#,
+  )?;
+
+  let mut cmd = Command::cargo_bin("parcel_css")?;
+  cmd.arg(file.path());
+  cmd.arg("--custom-media");
+  cmd.assert().success().stdout(predicate::str::contains(indoc! {r#"
+    @custom-media --foo print;
+  "#}));
+
```
devongovett, line 385: For next time, you could also have added this as a unit test - they are slightly easier to write and live in src/lib.rs. This is also fine though!

devongovett (approved):
Thanks!

PR #263, opened 2022-08-28
Support async bundling with JS resolver

devongovett:
This is an alternative approach to #196. This approach doesn't require converting the whole bundler to async. Instead, it uses napi threadsafe functions in combination with channels to wait for the result from the JavaScript thread in the rayon thread.

This required a fork of threadsafe_function.rs from napi-rs to support getting the result of the JS function, which I hope to potentially contribute upstream. This also means there is no required JS wrapper to convert from a callback-based API to a promise-based one.

Still need to do some performance testing to see generally how much a JS resolver slows things down.

cc. @dgp1130

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,9 +273,9 @@ checksum = "fb4a24b1aaf0fd0ce8b45161144d6f42cd91677fd5940fd431183eb023b3a2b8"
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.2"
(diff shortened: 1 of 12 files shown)


devongovett:
Basic performance testing indicates that using a resolver with `fs.readFileSync` is about 15% slower than native, and using `fs.promises.readFile` is about 40% slower.

dgp1130:
Very cool stuff! Excited to play around with this implementation.

PR #264, opened 2022-08-29
fix: resolution = 1 cannot be omitted from `image-set()`

yisibl:
Before this PR, we would always remove it when resolution = `1x/1dppx`. This is not compliant with the specification and will not render in the Safari <= 16 .

Input:

```css
body {
  background-image: image-set("../images/bg.webp" 1x, "../images/bg.webp" 2x);
}
```

## Before:

```css
body {
  background-image: image-set("../images/bg.webp", "../images/bg.webp" 2x);
}
```

## After:

```css
body {
  background-image: image-set("../images/bg.webp" 1x, "../images/bg.webp" 2x);
}
```

In the future we need to add detection of `resolution <= 0`.
For example: `image-set(url(example.png) 0x)` is an invalid syntax.

Closes: https://github.com/parcel-bundler/parcel-css/issues/262

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12260,17 +12260,36 @@ mod tests {
 
   #[test]
   fn test_image_set() {
+    // Spec: https://drafts.csswg.org/css-images-4/#image-set-notation
+    // WPT: https://github.com/web-platform-tests/wpt/blob/master/css/css-images/image-set/image-set-parsing.html
+    // test image-set(<string>)
     minify_test(
       ".foo { background: image-set(\"foo.png\" 2x, url(bar.png) 1x) }",
-      ".foo{background:image-set(\"foo.png\" 2x,\"bar.png\")}",
+      ".foo{background:image-set(\"foo.png\" 2x,\"bar.png\" 1x)}",
     );
+
+    // test image-set(type(<string>))
     minify_test(
       ".foo { background: image-set('foo.webp' type('webp'), url(foo.jpg)) }",
-      ".foo{background:image-set(\"foo.webp\" type(\"webp\"),\"foo.jpg\")}",
+      ".foo{background:image-set(\"foo.webp\" 1x type(\"webp\"),\"foo.jpg\" 1x)}",
     );
     minify_test(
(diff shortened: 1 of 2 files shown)


devongovett:
Thanks. I think I misinterpreted the [spec](https://drafts.csswg.org/css-images-4/#typedef-image-set-option). It does say the resolution is optional and defaults to 1x, but the grammar is `[ <resolution> || type(<string>) ]` which means that either a resolution or a type is required. So theoretically, we could still omit `1x` when there is a `type`. Does that work in browsers?

yisibl:
Yes, in fact, omitting the resolution is valid in Firefox 106, only Safari is not([Demo](https://codepen.io/team/css-tricks/pen/GRWbOWz?editors=0100)). This means that Safari does not comply with the standard implementation, even though it removes the prefix. If we were to distinguish these cases in code, it would probably be more expensive to maintain and confusing for the user to know when `1x` would be lost.

The following CSS works in Firefox 106, but not Safari.

```css
background-image: image-set( url("https://assets.codepen.io/t-90/dog.jpg") );
background-image: image-set( "https://assets.codepen.io/t-90/dog.jpg" );

background-image: image-set( "https://assets.codepen.io/t-90/dog.jpg" type("image/jpg") );
background-image: image-set(linear-gradient(black, white));
background-image: image-set( linear-gradient(black, white) type("image/png") );
```

devongovett:
Hmm it looks like Safari doesn't support `type()` at all.

yisibl:
The implementation in Safari is a subset of the specification, and from my testing, the only syntax supported in Safari is `image-set([ <url> | <string> ] [<resolution>])`

Overall, the syntax is too overly complex and even supports infinite nesting, which is terrible. So, I think we should keep it simple to handle. I will submit a csswg issue later on to discuss the simplified syntax.

```css
.foo {
  background-image: image-set(
    image-set(
      image-set(
        image-set(
          linear-gradient(green, green)
        )
      )
    )
  );
}
```

devongovett (approved):

PR #266, opened 2022-09-01
Suppress `UnsupportedPseudoClassOrElement` warning if `name` starts with `"-"`

pfroud:
https://github.com/parcel-bundler/parcel-css/issues/198#issuecomment-1231978976


devongovett (approved):
thanks!

PR #267, opened 2022-09-01
fix: CSS-wide keywords and `none` in `@keyframes` cannot remove quotes

yisibl:
In the `@keyframes` rule, this is an invalid syntax if the `<keyframes-name>` is CSS-wide keywords or `none` keyword.

> `<keyframes-name>` = `<custom-ident>` | `<string>`

Before this PR, we would always remove the quotes, which would result in generating an invalid CSS if the quotes happened to be CSS-wide keywords.

```
CSS-wide keywords =  initial | inherit | unset | default | revert | revert-layer
```

## Before
```css
/* Input */
@keyframes "revert" {}
@keyframes "none" {}

/* Output */
@keyframes revert {}
@keyframes none {}
```

## After
```css
/* Input */
@keyframes "revert" {}
@keyframes "none" {}

/* Output */
@keyframes "revert" {}
@keyframes "none" {}
```

## Throw an error

With the new parsing method, the following cases will throw an error, which was previously considered a legal CSS rule.

```css
/* Input */
@keyframes revert {}
@keyframes none {}

/* Output */
Unexpected token Ident("revert")
Unexpected token Ident("none")
```


src/rules/keyframes.rs
```diff
@@ -33,6 +35,67 @@ pub struct KeyframesRule<'i> {
...
+      KeyframesName::Ident(ident) => {
+        dest.write_ident(ident.0.as_ref())?;
+      }
+      KeyframesName::Custom(s) => {
+        // CSS-wide keywords and `none` cannot remove quotes.
+        match_ignore_ascii_case! { &*s,
+          "none" | "initial" | "inherit" | "unset" | "default" | "revert" | "revert-layer" => {
+            dest.write_char('"')?;
+            dest.write_str(s.as_ref())?;
+            dest.write_char('"')?;
```
devongovett, line 87: You should use the `serialize_string` function from the `cssparser` crate for this, as it will handle escaping quotes and such.

src/rules/mod.rs
```diff
@@ -259,7 +260,10 @@ impl<'i> CssRuleList<'i> {
     for mut rule in self.0.drain(..) {
       match &mut rule {
         CssRule::Keyframes(keyframes) => {
-          if context.unused_symbols.contains(keyframes.name.0.as_ref()) {
+          if context
+            .unused_symbols
+            .contains(&keyframes.name.to_css_string(Default::default()).unwrap())
```
devongovett, line 265: This will result in the name being quoted/escaped, but the `unused_symbols` list won't be. You'll need to do a `match` on the keyframe name to pull out the embedded string.

devongovett (commented):
Thanks! We'll need to update `AnimationName` in properties/animation.rs the same way as `KeyframesName`.

src/rules/keyframes.rs
yisibl: Done

yisibl:
> We'll need to update `AnimationName` in properties/animation.rs the same way as `KeyframesName`.

Can I fix it in the next PR? Here I want to focus on `@keyframes`.

PR #269, opened 2022-09-03
fix: added types for wasm init default export

deckchairlabs:
I noticed while using the wasm module within Deno, the `init` function wasn't properly typed. This PR fixes that.


devongovett (approved):
Thanks!

PR #278, opened 2022-09-09
Change link to `css-minimzer-webpack-plugin`

LekoArts:
With https://github.com/webpack-contrib/css-minimizer-webpack-plugin/pull/192 the link changed. Also updated the link text to be more accessible than "here"


devongovett (approved):
Thanks!

PR #285, opened 2022-09-14
Fix og:image

38elements:
This changed `og:image`, since `og:image` is url, not path.
https://ogp.me/


devongovett:
Ah I think I messed up the deploy. This has to be done by setting the `--public-url` during the build so that the image is actually copied. Should be fixed now.

PR #294, opened 2022-09-17
feat: support border-spacing

yisibl:
`border-spacing = <length> <length>?`

Spec: https://www.w3.org/TR/CSS22/tables.html#propdef-border-spacing

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -258,6 +258,69 @@ mod tests {
   );
 
   #[test]
+  pub fn test_border_spacing() {
+    minify_test(
+      r#"
+      .foo {
+        border-spacing: 0px;
+      }
+    "#,
+      indoc! {".foo{border-spacing:0}"
+      },
+    );
+    minify_test(
+      r#"
+      .foo {
+        border-spacing: 0px 0px;
+      }
+    "#,
+      indoc! {".foo{border-spacing:0}"
+      },
+    );
+
+    minify_test(
+      r#"
+      .foo {
+        border-spacing: 12px   0px;
+      }
+    "#,
+      indoc! {".foo{border-spacing:12px 0}"
+      },
+    );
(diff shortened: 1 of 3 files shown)


src/properties/mod.rs
```diff
@@ -805,6 +805,8 @@ define_properties! {
   "inset-inline": InsetInline(InsetInline) shorthand: true,
   "inset": Inset(Inset) shorthand: true,
 
+  "border-spacing": BorderSpacing(BorderSpacing),
```
devongovett, line 808: I think this could just be `Size2D<Length>` rather than needing a custom `BorderSpacing` type. `Size2D` already handles the parsing and omitting of the second value if it is equal to the first.

yisibl: 😂😂 Done

src/lib.rs
```diff
@@ -257,6 +257,73 @@ mod tests {
...
+    "#,
+      indoc! {".foo{border-spacing:6px 8px}"
+      },
+    );
+
+    // TODO: The `<length>` in border-spacing cannot have a negative value, 
+    // we may need to implement NonNegativeLength like Servo does.
+    // Servo Code: https://github.com/servo/servo/blob/08bc2d53579c9ab85415d4363888881b91df073b/components/style/values/specified/length.rs#L875
+    // CSSWG issue: https://lists.w3.org/Archives/Public/www-style/2008Sep/0161.html
+    // `border-spacing = <length> <length>?`
```
yisibl, line 315: @devongovett Do we need to implement `NonNegativeLength`?

devongovett: Eventually maybe. Right now it wouldn't have any effect on the output because we fall back to passing through the original value when an error occurs.

devongovett (approved):

PR #295, opened 2022-09-17
fix: avoid parsing `animation: "none"` to `animation: "none" none`.

yisibl:
In the `animation` property, special handling is required when the values of both `animation-name` and `animation-fill-mode` are `none`.

## Input

```css
.foo {
    animation: "none";
}

.bar {
  animation: "none" 2s;
}
```

## Before

```css
.foo {
  animation: "none" none;
}

.bar {
  animation: "none" 2s none;
}
```

## After

```css
.foo {
  animation-name: "none";
}

.bar {
  animation: "none" 2s;
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8196,7 +8196,10 @@ mod tests {
     minify_test(".foo { animation-name: \"test\" }", ".foo{animation-name:test}");
     minify_test(".foo { animation-name: foo, bar }", ".foo{animation-name:foo,bar}");
     minify_test(".foo { animation-name: \"none\" }", ".foo{animation-name:\"none\"}");
-    minify_test(".foo { animation-name: \"none\", foo }", ".foo{animation-name:\"none\",foo}");
+    minify_test(
+      ".foo { animation-name: \"none\", foo }",
+      ".foo{animation-name:\"none\",foo}",
+    );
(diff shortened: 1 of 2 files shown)


devongovett (approved):

devongovett:
Thanks. I updated the logic slightly in df280dc12747c18fd61b166498f216bbbe8662f6 and b6eec64432a935daab515af4130f84d8d9c8d88b, so that we serialize the name as a quoted string within the `animation` shorthand whenever it conflicts with a keyword from one of the other properties.

yisibl:
@devongovett Nice work, but what is the reason for serializing the animation-name to the end?

Also, we might want to consider browsers that don't support `animation-name: <string>`, like Chrome is implementing in: https://chromium-review.googlesource.com/c/chromium/src/+/3865188

devongovett:
Moving to the end fixed tests like this:

```css
animation: 2s alternate reverse
```

In this case, alternate is the direction, and reverse is the name. Without quotes, it's ambiguous so order is important.

Hadn't considered that strings were a new syntax though. Might need to update this...

devongovett:
Updated: 6e47cdb3c6050ba8d3ef51d55534716fcf3babc8

PR #296, opened 2022-09-17
fix: if the browser doesn't support `#rrggbbaa` color syntax, it is converted to `transparent`

yisibl:
Currently only enabled when `minify: true`.

```js
let opts = {
  minify: true,
  targets: {
    chrome: 61 << 16,
  }
}
```

## Input

```css
.foo {
  background-color: transparent;
  color: rgba(0, 0, 0, 0);
  border-color: rgb(0 0 0 / 0%);
  background: linear-gradient(90deg, rgba(0, 0, 0, 0), transparent);
}
```

## Before

```css
.foo{color:rgba(0,0,0,0);background:linear-gradient(90deg,rgba(0,0,0,0),rgba(0,0,0,0));border-color:rgba(0,0,0,0)}
```

## After

This is consistent with [cssnano](https://cssnano.co/playground/).

```css
.foo{color:transparent;background:linear-gradient(90deg,transparent,transparent);border-color:transparent}
```

### DIFF
--- a/src/values/color.rs
+++ b/src/values/color.rs
@@ -370,24 +370,30 @@ impl ToCss for CssColor {
           // If the #rrggbbaa syntax is not supported by the browser targets, output rgba()
           if let Some(targets) = dest.targets {
             if !Feature::CssRrggbbaa.is_compatible(targets) {
-              dest.write_str("rgba(")?;
-              write!(dest, "{}", color.red)?;
-              dest.delim(',', false)?;
-              write!(dest, "{}", color.green)?;
-              dest.delim(',', false)?;
-              write!(dest, "{}", color.blue)?;
-              dest.delim(',', false)?;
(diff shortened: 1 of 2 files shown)


devongovett (approved):

src/values/color.rs
```diff
@@ -370,24 +370,30 @@ impl ToCss for CssColor {
           // If the #rrggbbaa syntax is not supported by the browser targets, output rgba()
           if let Some(targets) = dest.targets {
             if !Feature::CssRrggbbaa.is_compatible(targets) {
```
yisibl, line 372: @devongovett Maybe it would be better to convert only when `dest.minify = true` here too? I don't really like the parser conversion format if the CSS is not compressed. What do you think?

devongovett: Sure

PR #304, opened 2022-09-23
feat: implement CSS columns

yisibl:
- Spec: https://www.w3.org/TR/css-multicol-1/#columns

```bash
  columns = <'column-width'> || <'column-count'>
```


yisibl:
@devongovett  PTAL

src/values/length.rs
```diff
@@ -99,6 +193,10 @@ macro_rules! define_length_units {
         $(#[$meta])*
         $name(CSSNumber),
       )+
+
+      /// A [`<length>`](https://www.w3.org/TR/css-values-4/#lengths) value
+      /// without a unit specified.
+      Unitless(CSSNumber),
```
devongovett, line 199: What is this needed for? Aren't unitless lengths only allowed for zero, and compute to pixels? So why not just store as pixels?

src/values/length.rs
```diff
@@ -35,6 +35,100 @@ impl LengthPercentage {
...
+}
+
+/// Either a [`<length>`](https://www.w3.org/TR/css-values-4/#lengths), or the `auto` keyword.
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(
+  feature = "serde",
+  derive(serde::Serialize, serde::Deserialize),
+  serde(tag = "type", content = "value", rename_all = "kebab-case")
+)]
+pub enum LengthOrAuto {
```
devongovett, line 92: We could probably make a generic type that allows "auto" or a type parameter. Happy to implement that separately.

src/properties/columns.rs
```diff
@@ -0,0 +1,109 @@
...
+    count: ColumnCount(IntegerOrAuto),
+  }
+}
+
+impl<'i> Parse<'i> for Columns {
+  fn parse<'t>(input: &mut Parser<'i, 't>) -> Result<Self, ParseError<'i, ParserError<'i>>> {
+    let state = input.state();
+    let with_or_count = LengthOrAuto::parse(input)?;
+
+    if let LengthOrAuto::Length(Length::Value(LengthValue::Unitless(..))) = with_or_count {
```
devongovett, line 32: Rather than parsing as a unitless length, you could instead first try parsing the column count, and if that fails, try parsing the length. When parsing shorthands where order doesn't matter, it can be useful to structure the code using a `loop`, trying to parse each property until values are found. There are many examples, [here's one](https://github.com/parcel-bundler/lightningcss/blob/865270134da6f25d66ce4a4ec0e4c4b93b90b759/src/properties/flex.rs#L81).

PR #324, opened 2022-10-29
feat(cli): Load browserslist configuration if no `--targets` supplied.

### LINKED ISSUE #281: Read `.browserslistrc` file from Rust binary when no targets
Skimming the code, I didn't see any files being read for this setting. The documentation talks about leaving a browserslist in the `package.json` but the Rust binary doesn't require Node and by proxy a `package.json`. It seems Parcel v2 [uses these RC files](https://github.com/parcel-bundler/parcel/blob/fe6158e5c728398bc618371dfbc12c6740ac9cc1/packages/core/core/src/requests/TargetRequest.js#L479). Also not opposed to manually passing to the config file (in lieu of assuming `$PWD/.browserslistrc`)--it would just be more convenient to keep the settings in a separate file (which could also just be an `lightningcssrc.{dhall,json,toml,yml}` file).

### LINKED ISSUE #293: Use browserslist entry in package.json with lightningcss-cli
When using the lightningcss-cli, the only way to provide a browserslist target seems to be the --target option, while the browserslist entry in package.json (and also a .browserslistrc file) is beeing ignored.  This means having to duplicate the browserslist target, but it would be much nicer if the lightningcss-cli would use the single definition in package.json like many other tools. The --target option could override the entry in package.json.

LeoniePhiline:
Fixes #293
Fixes #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,46 @@ To see all of the available options, use the `--help` argument:
 npx lightningcss --help
 ```
 
+#### Browserslist configuration
+
+If no `--targets` option is provided, then `lightningcss` finds browserslist configuration,
+selects queries by environment and loads the resulting queries as targets.
+
+Configuration resolution is modeled after the original `browserslist` nodeJS package.
+The configuration is resolved in the following order:
+
+- If a `BROWSERSLIST` environment variable is present, then load targets from its value. This is analog to the `--targets` CLI option.
+  _Example:_ `BROWSERSLIST="firefox ESR" lightningcss [OPTIONS] <INPUT_FILE>`
+- If a `BROWSERSLIST_CONFIG` environment variable is present, then resolve the file at the provided path.
+  Then parse and use targets from `package.json` or any browserslist configuration file pointed to by the environment variable.
+  _Example:_ `BROWSERSLIST_CONFIG="../config/browserslist" lightningcss [OPTIONS] <INPUT_FILE>`
+- If none of the above apply, then find, parse and use targets from the first `browserslist`, `package.json`
+  or `.browserslistrc` configuration file in any parent directory.
+
+Browserslist configuration files may contain sections denoted by angular brackets `[]`.
+Use these to specify different targets for different environments.
+Targets which are not placed in a section are added to `defaults` and used if no section applies matches the environment.
+
+_Example:_
+
+```
+# Defaults, applied when no other section matches the provided environment.
+firefox ESR
+
+[staging]
+# Targets applied only to the staging environment.
+samsung >= 4
+```
+
+When using parsed configuration from `browserslist`, `package.json` or `.browserslistrc` configuration files,
+the environment determined by
+
+- the `BROWSERSLIST_ENV` environment variable if present,
+- otherwise the `NODE_ENV` environment variable if present,
+- otherwise `production` is used.
+
+If no targets are found for the resulting environment, then the `defaults` configuration section is used.
+
 ### Error recovery
 
 By default, Lightning CSS is strict, and will error when parsing an invalid rule or declaration. However, sometimes you may encounter a third party library that you can't easily modify, which unintentionally contains invalid syntax, or IE-specific hacks. In these cases, you can enable the `errorRecovery` option (or `--error-recovery` CLI flag). This will skip over invalid rules and declarations, omitting them in the output, and producing a warning instead of an error. You should also open an issue or PR to fix the issue in the library if possible.
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ An extremely fast CSS parser, transformer, and minifier written in Rust. Use it 
   - Removing default property sub-values which will be inferred by browsers.
   - Many micro-optimizations, e.g. converting to shorter units, removing unnecessary quotation marks, etc.
 - **Vendor prefixing** – Lightning CSS accepts a list of browser targets, and automatically adds (and removes) vendor prefixes.
+- **Browserslist configuration** – Lightning CSS supports opt-in browserslist configuration discovery to resolve browser targets and integrate with your existing tools and config setup.
 - **Syntax lowering** – Lightning CSS parses modern CSS syntax, and generates more compatible output where needed, based on browser targets.
   - CSS Nesting (draft spec)
   - Custom media queries (draft spec)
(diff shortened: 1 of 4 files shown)


LeoniePhiline:
@devongovett 

This PR introduces browserslist configuration auto-discovery. ⚡

 The changes are pretty straightforward, as [`browserslist-rs` already perfectly mirrors](https://docs.rs/browserslist-rs/latest/src/browserslist/config/mod.rs.html) the nodeJS [`browserslist` behavior](https://github.com/browserslist/browserslist/blob/main/node.js#L226-L241).

I chose to require a negative boolean flag `--disable-browserslist` to facilitate the former `targets = None` setup.

This allows the [preserve_custom_media CLI integration test](https://github.com/parcel-bundler/lightningcss/commit/6efa74d77ec1ec3bc1ea332fa1495dca2ed6b661#diff-78a13ebd1717beafeb706c9df8c6296f5ae301dca5de530e668c772d9910449fR369-R387) to pass with an added `--disable-browserslist` option, while being the most ergonomic for basically everyone else:

I am convinced that the vast majority of users is going to want to use browserslist - with either configuration discovery or at least the built-in defaults. 
Therefore, providing the possibility to opt out appears to be the most beneficial choice, as compared to requiring almost all users to opt in.

However, this could be considered a breaking change needing a major semver bump.

If you prefer to stick to `1.x` right now, then I will change the browserslist auto-config feature from opt-out to opt-in by requiring a `--browserslist` flag, without which the CLI would exhibit the same behavior as before.

Please let me know what you prefer!

Detail question: 
Do you want `--(disable-)browserslist` and `--targets` to share an [`ArgGroup`](https://docs.rs/clap/3.2.2/clap/struct.ArgGroup.html) and be (enforced) mutually exclusive?

src/main.rs
```diff
@@ -46,6 +46,8 @@ struct CliArgs {
   #[clap(short, long, value_parser)]
   targets: Vec<String>,
   #[clap(long, value_parser)]
+  disable_browserslist: bool,
```
devongovett, line 49: Can we make this `--no-browserslist`? I think that naming convention for CLI flags is more common.

README.md
```diff
@@ -197,6 +197,48 @@ To see all of the available options, use the `--help` argument:
...
+```
+
+When using parsed configuration from `browserslist`, `package.json` or `.browserslistrc` configuration files,
+the environment determined by
+
+- the `BROWSERSLIST_ENV` environment variable if present,
+- otherwise the `NODE_ENV` environment variable if present,
+- otherwise `production` is used.
+
+If no targets are found for the resulting environment, then the `defaults` configuration section is used.
```
devongovett, line 240: This part might be a breaking change. I think loading a browserlist file is ok to enable by default, but if a project had no browserslist configuration, then it might be weird if we suddenly started compiling differently. I think opting into downleveling, either via a `--targets` option or via a browserslist config file is the right way to go.

LeoniePhiline:
Thank you for your review!

I will switch from opt out to opt in to keep the change non-breaking.

I'll also need to write some CLI integration tests for browserslist config discovery and browserslist defaults (where opted in).

src/main.rs
LeoniePhiline: I'll turn it into `--browserslist` to keep the feature opt-in; avoiding the breaking change of applying browserslist default targets.

For a future lightningcss 2.0 we could reconsider enabling browserslist by default.

README.md
LeoniePhiline: I will check if browserslist-rs provides an API to disable defaults. As far as I remember, it is rather locked down, as `config::load()` is not public. In that case, the defaults alone could be opt-in (e.g. `--browserslist-defaults`?) while 

Otherwise, I'll turn it into `--browserslist` to keep the feature opt-in; avoiding the breaking change of applying browserslist default targets.

For a future lightningcss 2.0 we could reconsider enabling browserslist by default.

devongovett: I think `--targets "defaults"` should already cover that case, no?

LeoniePhiline:
# Latest changes ([cb9a3e6](https://github.com/parcel-bundler/lightningcss/pull/324/commits/cb9a3e6d6d0c73bffe860b689a6f1936c481434f))

- The feature is opt-in via the `--browserslist` option, as not to break existing setups.
- The options `--browserslist` and `--targets <TARGET>` are now mutually exclusive, as enforced by `clap`.
- Browserslist integration is advertised more prominently than before, in the `README.md` features list.
- CLI integration tests were added, testing and documenting the various supported ways of configuration provision and discovery.

## Observations these changes were based on

### Browserslist config auto-discovery's unwanted effect on tests, as well as potential CLI users:

- In CLI integration tests, where no `--targets` option was used, specifically the `preserve_custom_media` test, the browserslist config discovery found and read from lightningcss' own `package.json` file, containing a `browserslist` section. I had initially had the misconception that this test failed because browserslist defaults were applied, but later realized the configuration had been read from [`package.json`](https://github.com/parcel-bundler/lightningcss/blob/master/package.json) .
- This effect could apply to many unsuspecting users: They currently run `lightningcss` without `--targets` option, yet they do have browserslist configuration in CWD or an ancestor directory - used by other tools in the same project or even for personal use in their `$HOME` directory. Enabling the browserslist config discovery feature by default would mean to silently change these users' transpilation output. __Therefore browserslist config discovery *must* be introduced as opt-in, lest it be a breaking change.__
- The `--browserslist` option must be advertised, as it must be enabled manually before use.

### Caveats in tests where browserslist config auto-discovery is enabled:

- To circumvent the "accidental config discovery" issue in our tests, whereever we want to have `--browserslist` enabled, we must set the working directory for CLI integration tests to be a temporary directory (as created by `assert_fs`) rather than the project root directory. 
- This is not bullet proof: Commonly the temporary directory's ancestor directories we have no power over. E.g. as `assert_fs::NamedTempFile::new()` creates a randomly named temporary folder like `/tmp/.tmpx1LOGu/`, browserslist might discover configuration files in its parent folder `/tmp/` if there happen to be such files for any possible reason. In Windows environments, the "Temp" folder might be located within `%USERPROFILE%` and have a long list of ancestor directories possibly containing browserslist configuration.
- In CI, `/tmp` should be clean when running in a fresh container; thus tests will be reliable despite the above caveat.
- Another side effect inductor is environment variables: For CLI integration tests which enable the configuration auto-discovery code path we must clear the environment. (At least `BROWSERSLIST`, `BROWSERSLIST_CONFIG`, `NODE_ENV` and `BROWSERSLIST_ENV`.)
- You will see `cmd.current_dir()` and `cmd.env_clear()` used in all browserslist-enabled CLI integration tests.

devongovett (approved):
Thank you, this looks great!

LeoniePhiline:
Thanks @devongovett - this means a lot to me.

Issue #320, opened 2022-10-24
Media query range contexts are sometimes incorrectly downgraded


romainmenke:
```css
@media screen and (not (200px <= width < 500px)) {
	html {
		background-color: green;
	}
}

@media screen and (not (min-width: 200px) and (max-width: 499.999px)) {
	html {
		background-color: purple;
	}
}
```

I would expect html background color to be `purple` but it is `green`.

When adding `not` before a media feature with a range context the result is a list with `not` and `and` on the same level. `not (min-width: 200px) and (max-width: 499.999px)`

This is not allowed by the specification : https://www.w3.org/TR/mediaqueries-5/#media-conditions

correct transform :

```css
@media screen and (not ((min-width: 200px) and (max-width: 499.999px))) {
	html {
		background-color: purple;
	}
}
```

LeoniePhiline:
The parens around the [`<media-not>`](https://www.w3.org/TR/mediaqueries-5/#typedef-media-not) are even non-standard, as the root-level [<media-condition-without-or>](https://www.w3.org/TR/mediaqueries-5/#typedef-media-condition-without-or) uses no parens. (Only the [<media-in-parens>](https://www.w3.org/TR/mediaqueries-5/#typedef-media-in-parens) inside it.)

Correct would be:

```css
@media screen and not ((min-width: 200px) and (max-width: 499.999px)) {
	html {
		background-color: purple;
	}
}
```

romainmenke:
Both of these work just fine and both are standard :

```css
@media screen and not ((min-width: 200px) and (max-width: 499.999px)) {
	html {
		background-color: purple;
	}
}

@media screen and (not ((min-width: 200px) and (max-width: 499.999px))) {
	html {
		background-color: hotpink;
	}
}
```

The grammar is recursive.
- `<media-in-parens>` can contain a `<media-condition>`
- `<media-condition>` can contain a `<media-not>`
- `<media-condition>` can contain a `<media-in-parens>`

But you can not mix `and` and `not` at the same level.

LeoniePhiline:
Ah yes. :) True.

romainmenke:
Thank you @devongovett

Issue #331, opened 2022-11-06
`compat.rs`: Media query interval syntax incorrectly marked as globally unsupported


LeoniePhiline:
# Range vs interval syntax

Reading the above sources, it appears to me as if browsers make no difference between range (`(width < 500px)`) and interval (`(200px <= width < 500px)`) syntax, following [CSS Media Queries 4](https://www.w3.org/TR/mediaqueries-4/#mq-range-context) while lightningcss treats range syntax as supported and interval syntax as (incorrectly) unsupported.

# Supported by various browser distribs:

- Can I Use: https://caniuse.com/css-media-range-syntax
- MDN Compat: https://developer.mozilla.org/en-US/docs/Web/CSS/@media#browser_compatibility under "**Range syntax from Media Queries Level 4**"
- Compat data available: https://github.com/mdn/browser-compat-data/blob/main/css/at-rules/media.json#L1304-L1341 

# Marked as globally unsupported:

- `compat.rs` (generated): https://github.com/parcel-bundler/lightningcss/blob/master/src/compat.rs#L1211-L1213
- Hard-coded global incompatibility in generator: https://github.com/parcel-bundler/lightningcss/blob/1eccce556606ce80956c744e82d23d2405072358/scripts/build-prefixes.js#L229

# Interval syntax demo

https://jsfiddle.net/cwmefnq9/

devongovett:
Looks like this was fixed in be440efe8feeaab7351cf3a27d58a1f062bcabf0

PR #332, opened 2022-11-06
fix: Correctly enclose negated downleveled interval media queries in parens

LeoniePhiline:
Fixes https://github.com/parcel-bundler/lightningcss/issues/320

# Changes

### Targets and range syntax

I changed relevant tests in `lib.rs` from firefox 60 and 85 to firefox 62 and 63, as media feature range syntax was introduced with firefox 63.
This takes into account that lightningcss currently treats range syntax and interval syntax as two different features. See also https://github.com/parcel-bundler/lightningcss/issues/331 

It might be that some browsers first implemented range syntax without interval syntax, and added interval syntax later. However, this is not documented in the compatibility tables. Lightningcss considers interval syntax as not supported anywhere, even though, in reality, current Gecko and Blink (but not WebKit, apparently) do support interval syntax today.

As interval syntax is now (if this PR be merged) implemented by rendering a `MediaCondition::Operation` of two `MediaFeature::Range` features, the output depends on the range syntax target support. Therefore, there are now two versions of each interval media feature query test - one using range syntax (firefox >= 63) and one using traditional syntax (firefox <= 62).

### Test placement

I had been unsure where to put the new, initially failing test reproducing https://github.com/parcel-bundler/lightningcss/issues/320. Placing it right in the `media_query` module seemed to be the right thing, but there is otherwise only a test for `and()`. All other tests - including media query tests - appear to be located in `lib.rs`. Should my added tests from `media_query.rs` be migrated over into `lib.rs`?

## Complex implementation, benefits and possible alternatives

The implementation of the fix might appear somewhat contrived at first sight. Yet in terms of behavior it seems to me to be the cleanest solution.

The issue is that if the downleveling only happens in `impl<'i> ToCss for MediaFeature<'i>` (as before the PR), then we cannot know if the feature is negated, the trait contract does not offer passing further options, except for the printer `dest`.

#### A) [Add a field to Printer](https://github.com/parcel-bundler/lightningcss/pull/333)

The `Printer` struct does have an `in_calc`  boolean already - would it be the right thing to add an `in_negated_media_condition` field? This looks a little like abuse to me. However, is the implementation in this PR the better choice?

Adding `in_negated_media_condition` to `struct Printer` would be an option. Feels wrong, but less complex. 

#### B) [Always wrap expanded interval in parens in `impl<'i> ToCss for MediaFeature<'i>`](https://github.com/parcel-bundler/lightningcss/pull/334)

Any expanded media feature interval could be enclosed in parens.
However: This outputs unnecessary parens for non-negated queries. 

E.g. `@media (200px >= width >= 100px)` would result in `@media ((max-width: 200px) and (min-width: 100px))`, where `@media (max-width: 200px) and (min-width: 100px)` would be sufficient.

#### C) [Using the solution provided in the PR you are currently looking at](https://github.com/parcel-bundler/lightningcss/pull/332)

Complex code, but clean output.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6806,7 +6806,7 @@ mod tests {
         }
       "#},
       Browsers {
-        firefox: Some(60 << 16),
+        firefox: Some(62 << 16),
         ..Browsers::default()
       },
     );
@@ -6827,7 +6827,28 @@ mod tests {
         }
       "#},
       Browsers {
-        firefox: Some(85 << 16),
+        firefox: Some(62 << 16),
+        ..Browsers::default()
+      },
+    );
+
+    prefix_test(
+      r#"
+        @media (100px <= width <= 200px) {
+          .foo {
+            color: chartreuse;
+          }
+        }
+      "#,
+      indoc! { r#"
+        @media (width >= 100px) and (width <= 200px) {
+          .foo {
+            color: #7fff00;
+          }
+        }
+      "#},
+      Browsers {
+        firefox: Some(63 << 16),
         ..Browsers::default()
       },
     );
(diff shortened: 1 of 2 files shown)


devongovett:
Thanks for putting together all these options and sorry for my slow response. I decided to completely remove the `InParens` variant from the AST, as it was causing some issues. Instead, we detect when parentheses are needed during printing, which simplifies transforms because output is guaranteed to be correct. See #343.

Issue #357, opened 2022-12-11
Slow compile times / slimming down dependencies


chinedufn:
Hey!

## Background

I am looking to compile a dynamic library that does some CSS parsing.

- The host passes a CSS string to the css dylib.

- The css dylib parses the CSS and tracks some information about it internally.

- The css dylib returns a modified CSS string to the host.

---

I am looking to use an existing crate to handle this use case.

I am currently evaluating `lightningcss` and [`cssparser`](https://github.com/servo/rust-cssparser).

## The Problem

I started writing my simple `dylib` using `cssparser`, but quickly found that it was too low level for my needs.

After searching on `crates.io` I found `lightningcss`. It was the perfect level of abstraction. I whipped up my css parsing dylib in no time. Great.

A day later I noticed that my builds had gotten slower since switching out `cssparser` for `lightningcss`.

## Reproducing

Here's how to see the difference in compile time impact between `cssparser` and `lightningcss`.

https://github.com/chinedufn/rust-cssparser-and-lightningcss-compile-times

```
git clone git@github.com:chinedufn/rust-cssparser-and-lightningcss-compile-times.git
./run.sh
```

#### Example Output

```
rust-cssparser debug build timing
real 0m11.188s user 0m29.876s sys 0m3.621s

lightningcss debug build timing
real 0m40.468s user 2m40.552s sys 0m16.839s
```

## Potential Solutions

I looked through `lightningcss`'s `Cargo.toml` to see if any crates could be made optional.

Here are ones that immediately stood out:

```toml
# Stood out as deps that we could potentially make optional

serde = { version = "1.0.123", features = ["derive"] }
parcel_sourcemap = { version = "2.1.1", features = ["json"] }
rayon = "1.5.1"
dashmap = "5.0.0"

# I didn't look into what this is used for, but putting it in here since it depends on the
# heavy `syn` crate.
lightningcss-derive = "..."
```

---

Here are ones that could potentially be removed entirely:

```
# itertools is barely used and does not seem necessary
itertools = "0.10.1"
# Looks like we could easily copy/paste what we're using.
pathdiff = "0.2.1"

# I didn't look at all of the dependencies.. but it seems like there are a few that are barely used
# and could be good candidates for removing and just copy/pasting the tiny bit of code that
# we actually depend on.
```

---

Additionally, I wonder if we can turn `default-features = false` for `lightningcss-derive`'s syn dependency
and only enable exactly what we're using.

## Alternative Solutions

[...]

PR #358, opened 2022-12-11
Introduce "bundler" feature flag

chinedufn:
This commit adds a "bundler" feature flag to the lightningcss crate.

We put the "dashmap" and "rayon" dependencies behind this new flag.

Users that do not need the bundler can now disable the feature to save
on compile time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,14 @@ name = "lightningcss"
 path = "src/lib.rs"
 crate-type = ["rlib"]
 
+[features]
+default = ["bundler", "grid"]
+browserslist = ["browserslist-rs"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -26,6 +26,14 @@ name = "lightningcss"
 path = "src/lib.rs"
 crate-type = ["rlib"]
 
+[features]
```
chinedufn, line 29: Moved the feature flags up since they're commonly found above the dependencies section. 
When evaluating a crate I find it nice to be able to quickly see what features I can turn off.

Happy to move them back if you prefer!

devongovett (approved):

PR #360, opened 2022-12-13
Put all of serde behind feature flag

chinedufn:
This commit makes the "serde" dependency completely optional.


Cargo.toml
```diff
@@ -27,7 +27,7 @@ path = "src/lib.rs"
 crate-type = ["rlib"]
 
 [dependencies]
-serde = { version = "1.0.123", features = ["derive"] }
+serde = { version = "1.0.123", features = ["derive"], optional = true }
```
chinedufn, line 30: This is a breaking change.

We could make it non-breaking by enabling the `serde` feature by default, but that would also enable `small-vec/serde` and `cssparser/serde`, which are not currently enabled by default.

It looks like `lightningcss` is on an alpha train. 
Is a small breaking change here acceptable?

Unless `serde` is commonly used for some reason, I think it makes sense to have it be disabled by default.

chinedufn (commented):
Happy to make any changes that you want to see to get this merged.

Thank you!

Cargo.toml
```diff
@@ -61,7 +61,7 @@ default = ["grid"]
 browserslist = ["browserslist-rs"]
 cli = ["clap", "serde_json", "browserslist", "jemallocator"]
 grid = []
-serde = ["smallvec/serde", "cssparser/serde"]
+with-serde = ["serde", "smallvec/serde", "cssparser/serde"]
```
chinedufn, line 64: This is a breaking change.

We could make it non-breaking by enabling the `with-serde` feature by default and having a `serde = []` feature, but that would enable `small-vec/serde` and `cssparser/serde`, which are not currently enabled by default.

It looks like `lightningcss` is on an alpha train. 
Is a small breaking change here acceptable?

Unless `serde` is commonly used for some reason, I think it makes sense to have it be disabled by default.

---

(Using the feature flag name "serde" as `serde = ["serde", "smallvec/serde", "cssparser/serde"]` wasn't possible since you can't define a feature flag with the same name as a dep.)

devongovett:
We might need to split this into two features. At the moment, this will break the Node bindings which depend on the currently not feature flagged serde derived traits. The currently flagged ones are for additional serialization support, e.g. if you wanted to serialize the whole AST.

If you run `cargo build --package lightningcss_node` you can see the errors. I'd rather not enable all of serde for the node binding, just the parts you've added here which are needed so we can serialize errors and dependencies that get returned. Not sure what to call the two features though, haha.

chinedufn:
Got it thanks for explaining.

I timed builds with and without all of the derives enabled and enabling them all added quite a bit to build time (15+ seconds for me).

So, makes sense.

---

I've added a `nodejs = ["serde"]` flag. The following derives are enabled when the `"nodejs"` feature is enabled.

- BundleErrorKind
- CssModuleReference
- CssModuleExport
- Dependency
- ImportDependency
- UrlDependency
- SourceRange
- Location
- Error
- ErrorLocation
- ParseError
- SelectorError
- MinifyErrorKind
- PrinterErrorKind
- Browsers
- CowArcStr

All other serde derives are only enabled when the `"with-serde"` feature is enabled.

The `lighningcss_node` crate has the `"nodejs"` feature on `lightningcss` enabled by default.

chinedufn:
Anything I can do to help this land so that it doesn't collect merge conflicts?

Cargo.toml
devongovett: I was able to rename it back to serde by doing this:

```
serde = ["dep:serde", "smallvec/serde", "cssparser/serde"]
```

devongovett (approved):
Thanks!

PR #361, opened 2022-12-13
Disable syn default features in lightningcss-derive

chinedufn:
This disables the default syn features in lightningcss-derive's
Cargo.toml.

It also removes an unused syn feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -696,7 +696,6 @@ version = "1.0.0-alpha.35"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


devongovett (approved):

devongovett:
Unfortunately this seems to have broken `cargo build -p lightningcss-derive` which runs when publishing the crate. I'm not sure why it isn't broken when building all of lightningcss though - maybe something else is enabling those features. Anyway, I don't think this should make too much difference now that the "visitor" feature flag of lightningcss exists - lightningcss-derive shouldn't get pulled in at all.

chinedufn:
Ah yeah I should've tried to build the crate on its own. Sounds good to me, thanks.

PR #371, opened 2022-12-28
ci: refactor node addon build

Brooooooklyn:
The refactored build make the `aarch64-unknown-linux-gnu` and `armv7-unknown-linux-gnueabihf` targets compatible with `GLIBC_2.17`


.github/workflows/release.yml
```diff
@@ -158,13 +156,13 @@ jobs:
           RUST_TARGET: ${{ matrix.target }}
       - name: Build CLI
         run: |
-          cargo build --release --features cli --target ${{ matrix.target }}
+          yarn napi build --bin lightningcss --release --features cli --target ${{ matrix.target }}
```
devongovett, line 159: Probably shouldn't be `napi` here? This is building the CLI, which is a standalone binary.

Brooooooklyn: `napi` is able to build binary, and I use napi to cross link `armv7-unknown-linux-gnueabihf` target here

devongovett (approved):

PR #372, opened 2022-12-28
Add sourcemap feature

chinedufn:
This commit introduces the "sourcemap" feature.

Users that do not need sourcemaps can disable this feature and save
roughly 15% on compile times.

Related to https://github.com/parcel-bundler/lightningcss/issues/357

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,14 +27,15 @@ path = "src/lib.rs"
 crate-type = ["rlib"]
 
 [features]
-default = ["bundler", "grid", "nodejs"]
+default = ["bundler", "grid", "nodejs", "sourcemap"]
 browserslist = ["browserslist-rs"]
-bundler = ["dashmap", "rayon"]
+bundler = ["dashmap", "sourcemap", "rayon"]
 cli = ["clap", "serde_json", "browserslist", "jemallocator"]
 grid = []
+jsonschema = ["schemars", "serde", "parcel_selectors/jsonschema"]
 nodejs = ["dep:serde"]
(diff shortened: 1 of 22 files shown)


chinedufn:
Seeing around 15% savings in compile times on my machine when running `cargo clean && cargo build --no-default-features --features grid --timings`.

This PR gets `rkyv` out of the dependency tree for those that aren't using sourcemaps.
`syn` is still in there but future PRs will try to chip away at that.

## Before

image

## After

image

src/stylesheet.rs
```diff
@@ -233,7 +235,12 @@ where
     let project_root = options.project_root.clone();
     let mut printer = Printer::new(&mut dest, options);
 
-    printer.sources = Some(&self.sources);
+    #[cfg(feature = "sourcemap")]
+    {
```
chinedufn, line 239: Conditionally compiling expressions isn't stable, so needed to wrap this in a block.

Cargo.toml
```diff
@@ -29,12 +29,13 @@ crate-type = ["rlib"]
 [features]
 default = ["bundler", "grid", "nodejs"]
```
devongovett, line 30: I think we should probably add "sourcemap" to the default features. I guess the node bindings are getting it through "bundler" right now, though they kinda use it directly too even without the bundler.

chinedufn: Ok great. Just force pushed a commit with sourcemap as an explicitly default feature

devongovett (approved):
Thanks!

PR #383, opened 2023-01-05
fix: `<layer-name>` should be escaped

yisibl:
Spec: https://w3c.github.io/csswg-drafts/css-cascade-5/#typedef-layer-name

Closes: #382


devongovett (approved):
Thanks!

Issue #386, opened 2023-01-07
possible to use in stackblitz?


mayank99:
When running in stackblitz, I get this error when importing lightningcss:

```
Error: Cannot find module '../lightningcss.linux-x64-gnu.node'
Require stack:
- /home/projects/node-n6my9w/node_modules/lightningcss/node/index.js
    at Module._resolveFilename (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:6:217308)
    at Module._load (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:6:214847)
    at Module.require (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:6:218087)
    at i (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:6:415284)
    at _0x434ff9 (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:15:142782)
    at Object.eval (file:///home/projects/node-n6my9w/node_modules/lightningcss/node/index.js:21:22)
    at Object.function (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:15:143544)
    at Module._compile (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:6:219079)
    at Module._extensions..js (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:6:219743)
    at Module.load (https://noden6my9w-ec0u.w-credentialless.staticblitz.com/blitz.8dd2e1d89f7d2e6b43a2b4af08588e58ffd9a2ea.js:6:217769) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
[...]

PR #387, opened 2023-01-07
Support node in wasm package

mayank99:
As discussed in #386, will try to read from filesystem (using `fs`) if used in a node environment, and fallback to `fetch` for browser/deno environments.

### DIFF
--- a/wasm/index.mjs
+++ b/wasm/index.mjs
@@ -5,7 +5,7 @@ let wasm;
 
 export default async function init(input = url) {
   if (typeof input === 'string' || (typeof Request === 'function' && input instanceof Request) || (typeof URL === 'function' && input instanceof URL)) {
-    input = fetch(input);
+    input = fetchOrReadFromFs(input);
   }
 
   const { instance } = await load(await input, {
@@ -51,3 +51,12 @@ async function load(module, imports) {
     }
   }
 }
+
+const fetchOrReadFromFs = async (inputPath) => {
+  try {
+    const fs = await import('node:fs');
+    return fs.readFileSync(inputPath);
+  } catch {
+    return fetch(inputPath);
+  }
+};
--- a/wasm/index.mjs
+++ b/wasm/index.mjs
@@ -52,9 +52,9 @@ async function load(module, imports) {
   }
 }
 
-const fetchOrReadFromFs = async (inputPath) => {
+async function fetchOrReadFromFs(inputPath) {
   try {
-    const fs = await import('node:fs');
+    const fs = await import('fs');
     return fs.readFileSync(inputPath);
   } catch {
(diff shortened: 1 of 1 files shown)


shyim:
It would be nice if we could just pass the file to that function. So it's abstract and anyone can fetch the file from they want :)

LeoDog896:
Also considering the fact that it doesn't work on node or the browser anyways (#394).

mayank99:
okay, promoting this out of draft since there is demand for it

@devongovett do let me know if you would like to see a different api (based on the [suggestion](https://github.com/parcel-bundler/lightningcss/pull/387#issuecomment-1378352048) above maybe?).

wasm/index.mjs
```diff
@@ -51,3 +51,12 @@ async function load(module, imports) {
     }
   }
 }
+
+async function fetchOrReadFromFs(inputPath) {
+  try {
+    const fs = await import('fs');
+    return fs.readFileSync(inputPath);
```
devongovett, line 58: Would we need to use a different API for Deno? Or can Deno `fetch` a `file://` url?

LeoDog896: I do know that deno can instantiate streaming ([full docs](https://deno.land/manual@v1.12.2/getting_started/webassembly)), but I'm not sure about the consequences of deno's design are.

mayank99: well, i spun up a deno deploy playground and tried to `await import('fs')` and got this error (as expected)

```
TypeError: Relative import path "fs" not prefixed with / or ./ or ../ from "file:///src/main.ts"
```

after that, i wondered if it currently even works in deno so i tried to run the snippet from the [docs](https://lightningcss.dev/docs.html#from-deno-or-in-browser) and got a similar error:

```
Relative import path "napi-wasm" not prefixed with / or ./ or ../ from
```

i then got it to work after i added an [import map](https://deno.land/manual@v1.29.4/basics/modules/import_maps):
```json
{
  "imports": {
    "napi-wasm": "https://unpkg.com/napi-wasm"
  }
}
```

then i tested passing a full url to `init` and got it work

```js
import { serve } from 'https://deno.land/std@0.155.0/http/server.ts';
import init, { transform } from 'https://unpkg.com/lightningcss-wasm';

await init('https://unpkg.com/lightningcss-wasm/lightningcss_node.wasm');

let {code, map} = transform({
  filename: 'style.css',
  code: new TextEncoder().encode('.foo { color: red }'),
  minify: true,
});

const css = new TextDecoder().decode(code);

serve((req: Request) => new Response(css)); // .foo{color:red}
```

all of that to say, my current PR should be fine as long as the import map is added.

devongovett (approved):

PR #406, opened 2023-01-23
Visitor: allow short-circuiting on error

georgeclaghorn:
In [Classmate](https://github.com/georgeclaghorn/classmate), visitors call fallible Ruby methods. I’d like to stop traversal at the first error and propagate it back up to the caller.

I currently approximate that by storing the error on the visitor and wrapping each `Visitor` method’s body in `if self.error.is_none() { ... }`. It would be nicer if I could just return an error from a visitor method and let it bubble up. This PR makes that possible.

The JS bindings’ `JsVisitor` intentionally remains infallible to preserve existing behavior, where it continues traversal after an error, but it could adopt the new behavior if desired.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -91,7 +91,7 @@ pub fn derive_visit_children(input: TokenStream) -> TokenStream {
         let name = ident
           .as_ref()
           .map_or_else(|| Member::Unnamed(index.into()), |ident| Member::Named(ident.clone()));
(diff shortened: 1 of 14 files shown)


devongovett:
Thanks this looks like a good feature. Will look more in depth soon!

devongovett:
Thanks again and sorry for the delay in looking at this. I also updated the JS transformer to use the fallible visitor functions as well: 366d923ef1f5789749d3acfb7014ab24694b8038.

PR #410, opened 2023-01-30
WASM: Only use `import.meta.url` in fallback scenario

nicksrandall:
I'm currently using this in a "classic" (not module) worker and `import.meta.url` is not defined which is breaking the call to the `URL` constructor, which I don't even use because I send in the path the WASM module manually. 

This change uses `import.meta.url` and the `URL` constructor only in the fallback case when a user doesn't pass in a custom url.


devongovett:
Thanks!

PR #425, opened 2023-02-12
feat: implementing the base `@scope`

yisibl:
This PR adds support for the base @scope.

```
@scope {
  .foo {
    display: flex;
  }
}
```


devongovett:
Do you think we should put this behind a flag like we have for nesting since the spec is still draft and potentially changing? Also since it is not fully implemented yet.

yisibl:
Ok, I've added a flag to disable it by default.

But I can't run `yarn website:build` locally, can you take a look?

Issue #397, opened 2023-01-13
Whether to consider a flattened visitor ?


noyobo:
In the case

Input.css:

```css
.a:not(.b:not(.c:not(.d))) {
  color: red;
}
```

will add prefix to  class selector.

```js
{
  Selector(path) {
    return path.map((item) => {
      if(item.type === 'class') {
        item.name = `prefix_${item.name}`;
      }
      return item
    });
  }
}
```

Only match `.a` selector. 

matching all class selectors that need to do this is very tedious and there are many cases of css combinations.

```js
{
  Selector(selector) {
      function prefix(selectors) {
        return selectors.map((selector) => {
          if (selector.type === 'class') {
            selector.name = `prefix_${selector.name}`;
          } else if (selector.type === 'pseudo-class') {
            if (selector.kind == 'not') {
              selector.selectors = selector.selectors.map((selector) => {
                return prefix(selector);
              });
            }
          }
          return selector;
        });
      }
      return prefix(selector);
    }
}
```

Or is there another way to write it?

devongovett:
We could add `SelectorComponent` as a visitor function to allow visiting individual components. This would also allow handling nested selectors (e.g. in `:not()`).

PR #429, opened 2023-02-19
Add `Visitor::visit_selector_component`

georgeclaghorn:
References #397. Implements the Rust side of https://github.com/parcel-bundler/lightningcss/issues/397#issuecomment-1381323367:

> We could add `SelectorComponent` as a visitor function to allow visiting individual components. This would also allow handling nested selectors (e.g. in `:not()`).

I could particularly use the ability to visit nested selectors in Rust.

### DIFF
--- a/src/selector.rs
+++ b/src/selector.rs
@@ -1863,8 +1863,32 @@ impl<'i, T: Visit<'i, T, V>, V: Visitor<'i, T>> Visit<'i, T, V> for Selector<'i>
     visitor.visit_selector(self)
   }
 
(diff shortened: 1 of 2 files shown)


georgeclaghorn:
I’m no longer sure about this, and I've made do without it, so I'm going to close.

It seems like we would want to call `Visitor::visit_selector_list` for the selector lists inside pseudo-classes, but these aren’t `SelectorList`s to “keep `Component` small”:

https://github.com/parcel-bundler/lightningcss/blob/82bca744014bcfcdbafe3580ab07423ee110f724/selectors/parser.rs#L1235-L1237

Issue #421, opened 2023-02-10
No such file or directory when output-file contains directories that are not yet created


benoitchantre:
When we run lightningcss from the CLI with the `--output-file` parameter, an error occurs when some directories of the output-file are missing.

To prevent the error, theses directories must exist before running lightningcss.

Would it be possible to create the required directories when they are missing?

devongovett:
Seems reasonable.

PR #432, opened 2023-02-21
Support creating missing directories with `--output-file` option

lucasweng:
- Resolves https://github.com/parcel-bundler/lightningcss/issues/421


devongovett (approved):
Thanks!

Issue #90, opened 2022-02-17
Support multiple files with CLI


benoitchantre:
Would it be possible to support multiple files as input / output using the CLI?

PostCSS and ESbuild allow to use a glob pattern for the input and have arguments to set the output directory and extension.

Examples:
- `postcss src/**/*.css --base src --dir dist/css --ext .min.css `
- `esbuild src/**/*.js --bundle --outdir=dist/JS --out-extension:.js=.min.js`

A similar solution would be very useful as many projects have more than one stylesheet.

devongovett:
Yep, this would be a nice feature. 👍

mrkldshv:
Hey! I'd like to take a stab at this issue. Do we want to use any crate for glob pattern?

mryechkin:
Would love to see this!

makkabi:
+1, I tried to replace postcss with lightningcss, but I'm missing three things: 1. Support for browserslist in package-json (use --target instead) 2. No built-in watch-mode (I use the onchange package for this) and 3. this (handle all css files in a folder), and this is what prevents me using lightningcss-cli for now, because the only workaround I see is one npm script for every css file, and that is hard du manually manage and will bloat the package.json file.

devongovett:
In the meantime, you can use bash scripting or something like the `find` command to process multiple files:

```bash
 find . -name '*.css' -exec lightningcss {} -o dist/{} \;
```

makkabi:
> In the meantime, you can use bash scripting or something like the `find` command to process multiple files:
> 
> ```shell
>  find . -name '*.css' -exec lightningcss {} -o dist/{} \;
> ```

Thanks! If I would really need it, I would try it, but for now I just stick to PostCSS. Btw., I saw that (as of now) other developers just assume that *.css works: [11ty-Sass-Skeleton by @5t3ph](https://github.com/5t3ph/11ty-sass-skeleton/commit/16d8a06838c2f46dedf15be971d2a229dc216a82)
(And it does work - until you have more than one css file...)

devongovett:
that's fine, but just wanted to point out that this is possible since your original comment implied that it was not:

> this is what prevents me using lightningcss-cli for now

LeoniePhiline:
@mrkldshv Are you still working on this?

mrkldshv:
Nope, feel free to take it!

LeoniePhiline:
Unfortunately I also won't manage currently to work on this. Free to take!

[...]

PR #434, opened 2023-02-22
feat: support usage of globs in cli mode

thepassle:
**EDIT:** I seemed to have fixed the issue 🙂 The PR can be reviewed now

image
🎉

~~this PR isnt quite finished, it seems to be almost working, but I'm running into one issue that I cant quite figure out, I was hoping you could give me some pointers on that one, @devongovett ?~~

~~The issue is the following:~~

```
error[E0308]: `if` and `else` have incompatible types
   --> src/main.rs:252:11
    |
249 |           let css_modules_filename = if let Some(name) = css_modules {
    |  ____________________________________-
250 | |           name
    | |           ---- expected because of this
251 | |         } else {
252 | |           infer_css_modules_filename(&output_file)?
    | |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |           |
    | |           expected `&std::string::String`, found struct `std::string::String`
    | |           help: consider borrowing here: `&infer_css_modules_filename(&output_file)?`
253 | |         };
    | |_________- `if` and `else` have incompatible types

For more information about this error, try `rustc --explain E0308`.
error: could not compile `lightningcss` due to previous error
```

fixes https://github.com/parcel-bundler/lightningcss/issues/90

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -247,7 +247,7 @@ pub fn main() -> Result<(), std::io::Error> {
   
       if let Some(css_modules) = &cli_args.css_modules {
         let css_modules_filename = if let Some(name) = css_modules {
-          name
+          name.to_string()
         } else {
           infer_css_modules_filename(&output_file)?
         };
--- a/src/main.rs
+++ b/src/main.rs
@@ -67,7 +67,6 @@ struct SourceMapJson<'a> {
   names: &'a Vec<String>,
 }
 
-#[derive(Debug)]
 struct InputFile {
   filename: String,
   source: String
@@ -81,8 +80,6 @@ pub fn main() -> Result<(), std::io::Error> {
 
   match &cli_args.input_file {
     Some(g) => {
-      let globs = glob(g);
-
       match glob(g) {
         Ok(file_paths) => {
           for file in file_paths {
@@ -113,12 +110,9 @@ pub fn main() -> Result<(), std::io::Error> {
       files.push(InputFile {filename, source: contents});
     }
   }
-
-  println!("{:#?}", &files);
   
   for file in files {
     let InputFile { filename, source } = file;
-    println!("{}, {}", &filename, &source);
 
     let css_modules = if let Some(_) = cli_args.css_modules {
       let pattern = if let Some(pattern) = cli_args.css_modules_pattern.as_ref() {
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,4 @@
+use glob::glob;
 use atty::Stream;
 use clap::{ArgGroup, Parser};
 use lightningcss::bundler::{Bundler, FileProvider};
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -66,26 +67,41 @@ struct SourceMapJson<'a> {
...
-      (filename, contents)
-    }
+  match &cli_args.input_file {
+    Some(g) => {
+      let globs = glob(g);
+
+      match glob(g) {
+        Ok(file_paths) => {
+          for file in file_paths {
+            if let Ok(f) = file {
```
thepassle, line 89: I was also wondering if you could give me some pointers on this code, there's quite a bit of nested pattern matching going on, is there a cleaner/more idiomatic way to approach this?

src/main.rs
```diff
@@ -94,157 +110,164 @@ pub fn main() -> Result<(), std::io::Error> {
...
+  }
 
-  let css_modules = if let Some(_) = cli_args.css_modules {
-    let pattern = if let Some(pattern) = cli_args.css_modules_pattern.as_ref() {
-      match lightningcss::css_modules::Pattern::parse(pattern) {
-        Ok(p) => p,
-        Err(e) => {
-          eprintln!("{}", e);
-          std::process::exit(1);
+  println!("{:#?}", &files);
```
thepassle, line 117: I'll clean these up later, I was using them to debug while developing locally 🙃

src/main.rs
devongovett: One question I have is if we need to expand globs ourselves here or not? Usually the shell (e.g. bash) will expand globs before executing the program, so the glob itself will never reach it, it'll just be a list of files.

thepassle: During development I used the following command to run the cli:
```
⚡ cargo run --features="cli" "*.css"
    Finished dev [unoptimized + debuginfo] target(s) in 0.58s
     Running `target/debug/lightningcss '*.css'`
html {
  color: red;
}

body {
  background: green;
}
```

Having the quotes around the glob does seem needed, if you provide the glob without quotes, it does indeed get expanded by the shell, and result in the following:

```
⚡ cargo run --features="cli" *.css
    Finished dev [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/lightningcss test.css test2.css`
error: Found argument 'test2.css' which wasn't expected, or isn't valid in this context

USAGE:
    lightningcss [OPTIONS] [INPUT_FILE]

For more information try --help
```

devongovett:
Hey sorry it took so long for me to get back to you on this. I think we will need to change a few things.

At the moment, this will take only glob literal, which you need to quote to prevent your shell from expanding them. It would be nice if you could do `lightningcss a.css b.css` too. Then we wouldn't need to implement globs ourselves either, because `lightningcss *.css` would be expanded by your shell into multiple files and we get glob support for free.

In addition, at the moment this will compile each file, but then output them all concatenated together to stdout. If you use the `-o` option to output to a file, they will overwrite each other so only the last one is output. We'll need a way to somehow output each file with a different filename, perhaps by specifying a directory and preserving the filename of each input file.

PR #439, opened 2023-03-04
website: Make title visible in selection

notzheng:
Now, website title text is invisible in selection cause it's color is transparent.

CleanShot 2023-03-04 at 22 03 21@2x

Override its color in `::selection` to fix this.

CleanShot 2023-03-04 at 22 06 25@2x

Also the title in docs.

CleanShot 2023-03-04 at 22 07 29@2x

### DIFF
--- a/website/docs.css
+++ b/website/docs.css
@@ -64,6 +64,11 @@ header .title {
   text-decoration: none;
 }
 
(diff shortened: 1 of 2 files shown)


devongovett (approved):
Thanks!

PR #442, opened 2023-03-11
Bump `phf` dependency in `parcel_selectors` crate

nicoburns:
## Context

We're wanting to [use lightningcss](https://github.com/DioxusLabs/taffy/pull/393) as a dependency in [Taffy](https://github.com/DioxusLabs/taffy). However our CI is [complaining](https://github.com/DioxusLabs/taffy/actions/runs/4394482038/jobs/7695555149) that `lightningcss` is depending on multiple versions of the same crate (which we try to avoid for binary size reasons).

I've tracked this down to `parcel_selectors` crate which depends on `phf 0.8`on `cssparser 0.29` which depends on `phf 0.10`.

## Changes made

- Updates the `parcel_selectors` crate to use `phf 0.10` (no code changes needed)
- Bumps the version of the `parcel_selectors` crate to from `0.25.1` to `0.25.2`
- Updates the `repository` field of `Cargo.toml` to point to this repo
- Updates the `documentatation` field of `Cargo.toml` to point to docs for the `parcel_selectors` crate rather than the `selectors` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.4",
+ "getrandom",
  "once_cell",
  "serde",
  "version_check",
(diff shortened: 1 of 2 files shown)


devongovett (approved):
Thanks!

nicoburns:
@devongovett Thanks for merging this. Would it be possible to publish the new version of the `parcel_selectors` crate containing this fix? I think that's all the should be required as it's a semver compatible release.

PR #465, opened 2023-04-14
Update relative color parsing to latest spec

### LINKED ISSUE #809: lowering for `oklch` does not work when using numeric or mixed values
Currently, lightningcss lowers the following correctly (albeit with an unnecessary conversion from `oklch` to `lab`):

```css
--color-1: oklch(100% none none);
--color-2: oklch(100% 0 0);
```

Whereas the following are ignored completely:

```css
--color-3: oklch(1 0 0);
--color-4: oklch(100% 0% 0deg);
```

See [playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Afalse%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A7208960%2C%22firefox%22%3A7208960%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22body%20%7B%5Cn%20%20--color-1%3A%20oklch(100%25%20none%20none)%3B%5Cn%20%20--color-2%3A%20oklch(100%25%200%200)%3B%5Cn%20%20--color-3%3A%20oklch(1%200%200)%3B%5Cn%20%20--color-4%3A%20oklch(100%25%200%25%200deg)%3B%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D).

I came across a similar issue (#445) which was closed by https://github.com/parcel-bundler/lightningcss/commit/835e5bc68da7ecc7c2e59b20617eed56465f9045, but it doesn't work in the latest version.

devongovett:
Closes #809

The CSS spec was updated in https://github.com/w3c/csswg-drafts/issues/7876 to treat many channel values as only numbers rather than percentages or numbers for the purposes of calculations in relative colors. This is a breaking change since it no longer allows syntax like `calc(l + 10%)`, you need to do `calc(l + 10)` or `calc(l * 1.1)` depending on what you want to do. In addition, the range of each channel is different, for example `l` is 0-100 in lch, but 0-1 in oklch.

Unfortunately, we did not ship relative color syntax under a draft flag, so if we release this change in a minor version it might be an unexpected breaking change. Haven't decided what to do here yet.


nicksrandall:
@devongovett what would it take to merge this? Tailwind v4 css (for example) ships `oklch` colors by default and they are not being lowered because they use the decimal syntax and not the percentage.

devongovett:
So, either we merge it and accept the breaking change as a "fix" since the spec itself had a breaking change, make it opt-in (not great for new projects), or wait for a major version of lightningcss.

nicksrandall:
I vote for shipping it as a fix. I can't imagine there are many people relying on this behavior.

nicksrandall:
On the flip side, I don't know that anybody would be bothered by a major version change that didn't actually change anything meaningful? That probably is the safest bet and, IMO, there is no real reason to *wait* for a major version?

devongovett:
I decided to just go ahead and merge this. If it breaks something, I blame the CSS spec authors. 😂

PR #485, opened 2023-05-07
Improve support for `CustomFunction` transforms

phoenix-ru:
This PR enables authoring custom transformers for `PseudoElement` and `PseudoClass` `CustomFunction`s.

## Parsing whitespace inside custom pseudo functions is now more conservative
Previously, `:foo(.bar .baz .qux)` was parsed as `:foo(.bar.baz.qux)` which greatly limited custom transformers.

This PR bypasses aggressive whitespace removal inside `TokenList::parse_into` behind a flag.

To ensure that no API change is introduced, conservative parsing is only enabled for `CustomFunction` via new `TokenList::parse_preserve_whitespace`.

## `Selector` manipulations got two more public methods
`Selector::append` is a useful method for adding extra components to selectors, however it is quite limited when adding `Component::Combinator` and makes it impossible to reliably append complex expressions like `.foo > .bar #baz`

- `Selector::insert_raw(&mut self, index: usize, component: Component)` can be used for inserting new selector components into an arbitrary location, bypassing the limitations of `append`;
- `Selector::insert_raw_multiple(&mut self, index: usize, mut components: Vec<Component>)` is an optimized version of `insert_raw` to avoid the extra cost of allocations and memcpy when making multiple calls.

## Tests
Tests were added to ensure the correctness of a different parsing strategy. No existing tests changed

### DIFF
--- a/examples/custom_at_rule.rs
+++ b/examples/custom_at_rule.rs
@@ -22,6 +22,7 @@ fn main() {
   let source = std::fs::read_to_string(&args[1]).unwrap();
   let opts = ParserOptions {
     filename: args[1].clone(),
+    nesting: true,
     ..Default::default()
   };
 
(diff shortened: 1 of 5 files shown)


phoenix-ru:
Based on the current Pull Request, I implemented Vue's scoped style transforms using `lightningcss`

https://github.com/phoenix-ru/rust-vue-compiler/blob/8c97a06289695bf6fec146431c882d1b175b263d/crates/fervid_css/src/lib.rs#L53-L147

devongovett:
I think this commit should cover this usecase as well: 6a2adb64a44f89de628699952bb639cebda48389. That will preserve all tokens inside pseudo selector functions unmodified.

PR #492, opened 2023-05-19
Removed non-existing `-moz-placeholder-shown`

avi12:
https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions#mozilla-only_properties
Screenshot of -moz-* prefixes


devongovett:
hmm, autoprefixer generates the `-moz-placeholder-shown` prefix. You can try it here https://autoprefixer.github.io

Issue #506, opened 2023-06-07
css variable declaration calc minification


MoritzLoewenstein:
Currently, the following calc expression will not be minified:
```css
:root {
    --foo: calc(10px + 20px);
}
```
I know that it would break js code that relies on the computed style value for that property.

I will consider trying to implement this if there is a chance that it will me merged (as an opt-in option),
otherwise feel free to close :)

devongovett:
It is sorta hard if there are any unknown parameters like variables in there (e.g. `calc(10px + var(--foo))`). Even simplifying things like `calc(10px + 20px + var(--bar))` might be challenging to do right. But for simple cases like the one you posted where all terms are known it could be done. If you're interested in contributing that, you could start in here and add a new handler for `"calc"` functions: https://github.com/parcel-bundler/lightningcss/blob/1203368b722ef0c9dffb03e70245423e352f6cff/src/properties/custom.rs#L380

From there, it would need to try parsing each possible type that supports `calc`, e.g. lengths, angles, times, etc. similar to this: https://github.com/parcel-bundler/lightningcss/blob/master/src/properties/custom.rs#L460-L470

The difference would be using `Length` instead of `LengthValue` since `Length` supports `calc` but `LengthValue` does not. Then the result would just be a simplified value that could be stored as a `TokenOrValue::Length` etc.

Let me know if you have questions!

MoritzLoewenstein:
Thanks for the explanation, started the implementation in #526. I will let you know when its ready or if I have questions!

PR #526, opened 2023-06-26
WIP: css variable calc minification

MoritzLoewenstein:
Im trying to implement calc minification for css variables (#506).


MoritzLoewenstein:
Possibly? If you want, you can try that approach, I was struggling to implement the more useful transforms because I dont have any rust experience anyways.

PR #528, opened 2023-07-03
Fix bundle options type

ArnaudBarre:
Found via https://github.com/vitejs/vite/pull/13701
Introduced in https://github.com/parcel-bundler/lightningcss/pull/524


devongovett:
Oh oops sorry about that

devongovett:
Actually I think omitting this was just wrong. Dependencies are still possible, just not `@import`. I'm going to revert that.

PR #535, opened 2023-07-09
docs: update vite usage docs

azat-io:
Vite since version 4.4.0 supports LightningCSS out of the box.


devongovett:
Thanks! We should also add a note about the minifier option in the Vite config since I believe they are configured separately. We could link to the section of the Vite docs too.

PR #564, opened 2023-08-17
Avoid const enum for Features type

ArnaudBarre:
`const enum` can't be used with the [isolatedModules flag](https://www.typescriptlang.org/tsconfig#isolatedModules) which  is required by any compiler that is not tsc like esbuild which is used to compile the Vite config.

The proposed type described better the actual runtime code (I'm using `;` because this is the default prettier format, feel free to change if that's an issue for you)

### DIFF
--- a/node/targets.d.ts
+++ b/node/targets.d.ts
@@ -12,28 +12,28 @@ export interface Targets {
   samsung?: number
 }
 
-export declare const enum Features {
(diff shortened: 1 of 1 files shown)


devongovett:
Thanks. Those types are generated by a script so I've gone ahead and updated that.

PR #574, opened 2023-08-29
feat: Make `Visitor` bounded by `?Sized`

kdy1:
It does not have to be `Sized`.

# Real usage

I'm trying to use `lightningcss` for some SWC-related projects and I need to create a trait object of `Visitor`

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -98,7 +98,7 @@ fn derive(
 
   impl_generics
     .params
-    .push(parse_quote! { #v: crate::visitor::Visitor<#lifetime, #t> });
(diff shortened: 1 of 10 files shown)


kdy1:
Oops I found that `const TYPES: VisitTypes;` is now the problem

devongovett:
@kdy1 I think we can remove that since there is a visit_types function instead. Most places seem to already use it. If we just make that function required instead it should be ok. The compiler should still be able to inline it if it is constant. https://github.com/parcel-bundler/lightningcss/blob/master/src/visitor.rs#L155

kdy1:
I did it

devongovett (approved):

PR #586, opened 2023-09-17
Support @scope rule

### LINKED ISSUE #566: Support `@scope` rule
### Describe the feature

[Chrome 118 is shipping soon](https://chromium-review.googlesource.com/c/chromium/src/+/4790561), which is expected to be one of the more popular CSS features, giving new hope for style isolation of components.

- Explainer https://css.oddbird.net/scope/explainer
- Specification https://drafts.csswg.org/css-cascade-6/#scope-atrule

### Example:

```html
<style>
  @scope (.foo) to (.limit) {
    .green { background-color: green; }
  }
</style>

<div class=foo>
  <div class=green>Green</div>
  <div class=limit>
      <div class=green>Not green (within .foo, but below .limit)</div>
  </div>
</div>

<div class=green>Not green (not within .foo)</div>
```

Authors can also automatically scope the styles to `<style>`’s parent element by dropping the selector(s) in `@scope`’s prelude:

```html
<div>
  <style>
    @scope {
      .green { background-color: green; }
    }
  </style>
  <div class=green>Green</div>
</div>

<div class=green>Not green</div>
```

devongovett:
Based on #425 but with support for scope start and scope end. Closes #566

### DIFF
--- a/node/src/transformer.rs
+++ b/node/src/transformer.rs
@@ -281,6 +281,7 @@ impl<'i> Visitor<'i, AtRule<'i>> for JsVisitor {
             CssRule::LayerStatement(..) => "layer-statement",
             CssRule::Property(..) => "property",
             CssRule::Container(..) => "container",
+            CssRule::Scope(..) => "scope",
             CssRule::MozDocument(..) => "moz-document",
             CssRule::Nesting(..) => "nesting",
             CssRule::Viewport(..) => "viewport",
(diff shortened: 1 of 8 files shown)


devongovett:
One question is how it should behave with nesting. I requested clarification. https://github.com/w3c/csswg-drafts/issues/9370 However I think any browser that supports `@scope` will also support nesting so maybe it doesn't matter too much for now.

PR #597, opened 2023-10-06
feat: Add `IntoOwned` trait

### LINKED ISSUE #596: `IntoOwned` as a trait, and implement it for nodes when `R` implements `IntoOwned`
@devongovett Will you accept a PR for this?
If so, which crate should have `IntoOwned`?

kdy1:
Closes #596

The signature should be

```rust
pub trait IntoOwned<'any> {
  type Owned: 'any;

  /// Make lifetime of `self` `'static`.
  fn into_owned(self) -> Self::Owned;
}

```

# Description

 - The caller determines the trait to call, so the lifetime `'any` of `IntoOwned<'any>` is determined by the caller.

This is identical as `fn into_owned<'x>(self) -> Self<'x>`.

 - The trait is implemented for nodes with any lifetime (not 'any), so it can be called if `self` is provided.

 - The bound for generics is `for<'aa> T: IntoOwned<'aa>`, so the implementation of generic parameters do not matter.

### DIFF
--- a/derive/src/into_owned.rs
+++ b/derive/src/into_owned.rs
@@ -1,3 +1,5 @@
+use std::f32::consts::E;
+
 use proc_macro::{self, TokenStream};
 use proc_macro2::Span;
 use quote::quote;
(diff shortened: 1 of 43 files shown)


kdy1:
@devongovett I did it 😄 

The code is not clean, though.

kdy1:
Feel free to modify docs or codes 👍

devongovett:
Thanks, this looks good. I'll play around with it in more detail tomorrow.

devongovett (approved):

PR #599, opened 2023-10-10
feat: Implement `IntoOwned` for more types

kdy1:
I forgot to add `derive`s.

But I'm stuck due to an API design issue.

Is it fine to implement `IntoOwned` for the selector types? If so, where should the trait definition go?

### DIFF
--- a/derive/src/into_owned.rs
+++ b/derive/src/into_owned.rs
@@ -152,7 +152,15 @@ pub(crate) fn derive_into_owned(input: TokenStream) -> TokenStream {
       }
     }
   } else {
-    let params = generics.type_params();
+    let mut generics_without_default = generics.clone();
+
+    for p in generics_without_default.params.iter_mut() {
+      if let syn::GenericParam::Type(ty) = p {
+        ty.default = None;
+      }
+    }
+
+    let params = generics_without_default.type_params();
     quote! {
       impl #impl_generics crate::traits::IntoOwned<'any> for #self_name #ty_generics #where_clause {
         type Owned = #self_name<'any, #(#params),*>;
--- a/derive/src/into_owned.rs
+++ b/derive/src/into_owned.rs
@@ -184,9 +184,7 @@ fn into_owned(ty: &Type, name: proc_macro2::TokenStream) -> proc_macro2::TokenSt
             _ => quote! { #v.into_owned() },
           };
           quote! { #name.map(|#v| #into_owned) }
-        } else if last.ident == "Vec"
-          || last.ident == "SmallVec"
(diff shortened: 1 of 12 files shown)


kdy1:
cc @devongovett

devongovett:
hmm good question. selectors are annoying. I started by trying to use the existing `selectors` crate but have needed to make more and more modifications. At some point I'll probably refactor it.

Seems like we have 4 options:

1. Implement `IntoOwned` for selector types manually within lightningcss.
2. Merge selectors into lightningcss and use derive.
3. Use derive within selectors crate, and modify it so it uses the trait from lightningcss.
4. Move traits into a separate crate and use derive.

(3) seems like it would be easy but seemed like you had to do some hacking to get derive to import `lightningcss` from within lightningcss. Is (4) better? Seems like a lot to add a whole new crate just for traits but maybe? Not sure. Any thoughts? I can experiment in a bit as well.

kdy1:
I think 4 would be rather easy if you are fine with a new crate. I agree that a new crate can be too much for this... but I think this crate can be generally useful.

devongovett:
Actually yeah, moving only IntoOwned to a separate crate is a pretty good idea since it isn't really lightningcss specific. If you want to do that within the repo for now, maybe I can move it to a separate one later on. Not really sure of a good name for it that's available on crates.io... Maybe just `into-owned-trait` (unfortunately `into-owned` is taken but unused). If you think of a better one though feel free.

Btw, since it's now a trait, I think a bunch of the special cases (e.g. Option and Box) can also be implemented with generics instead of in the derive macro so it might get a lot simpler.

kdy1:
What about `static-self`?

kdy1:
Okay... It seems like it's almost done.

devongovett:
> What about static-self?

I like it!

kdy1:
I did it 😄

kdy1:
Ah. I think the code for derive macro require some cleanup

devongovett:
Sure, yeah I can take a look. Thanks so much for working on this. It'll be really helpful for some use cases I have in mind too. I'll look into moving these crates into a separate repo at some point and I can add you as a maintainer there too.

kdy1:
@devongovett May I ask when a new version will be published?

devongovett (approved):

devongovett:
Done https://github.com/parcel-bundler/lightningcss/commit/938fc8c76a663b2d15420d145eb6505efd91f002

kdy1:
Thank you!

olishiz:
Whoa, that's cool I'm in a part of a Pull Request.

PR #643, opened 2023-12-08
feat: support CJS output for wasm build

kirjavascript:
* Brings WASM builds feature parity with exports from native builds
* Adds esbuild to convert to CJS (similar to other parcel repos)

### DIFF
--- a/scripts/build-wasm.js
+++ b/scripts/build-wasm.js
@@ -40,6 +40,7 @@ const cjsBuild = {
   bundle: true,
   format: 'cjs',
   platform: 'node',
+  packages: 'external',
   outdir: `${dir}/wasm`,
   outExtension: { '.js': '.cjs' },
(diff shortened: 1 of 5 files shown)


scripts/build-wasm.js
```diff
@@ -31,19 +32,40 @@ let readme = fs.readFileSync(`${dir}/README.md`, 'utf8');
 readme = readme.replace('# ⚡️ Lightning CSS', '# ⚡️ lightningcss-wasm');
 fs.writeFileSync(`${dir}/wasm/README.md`, readme);
 
+const cjsBuild = {
+  entryPoints: [
+    `${dir}/wasm/wasm-node.mjs`,
+    `${dir}/wasm/index.mjs`,
+  ],
+  bundle: true,
```
devongovett, line 40: I think we should exclude node_modules here so that napi-wasm doesn't get bundled.

PR #651, opened 2023-12-26
feat(target): add riscv64 support

phanen:


devongovett:
What hardware runs RISC-V? I don't know much about it, but I'm concerned that there isn't much real-world hardware support for this architecture and no way for me to test it.

PR #652, opened 2024-01-01
improve bench

romainmenke:
The current bench script is flawed:
- it's comparing apples and oranges
- the bench script only does a single run of each tool
- it uses sources which are known to favor `lightningcss`

I think it would be better to have a good set of benchmarks which compare similar tools.

To really compare `lightningcss` to something you would need :
- representative CSS, not just bootstrap
- compare it to other tools with similar capabilities

In this PR I compared it to `@csstools/postcss-minify` + `postcss-preset-env` on a different source and then I get these results:

```
┌─────────┬─────────────────────────────────────────────────┬──────────────┬──────────────┬──────────────┬───────────┬──────────────────────────┐
│ (index) │                      label                      │ percentile50 │ percentile95 │ percentile99 │    max    │ percentageOfOriginalSize │
├─────────┼─────────────────────────────────────────────────┼──────────────┼──────────────┼──────────────┼───────────┼──────────────────────────┤
│    0    │                    'cssnano'                    │   '6.4ms'    │   '12.1ms'   │   '17.6ms'   │ '142.8ms' │         '82.46%'         │
│    1    │           '@csstools/postcss-minify'            │   '0.8ms'    │   '3.0ms'    │   '6.4ms'    │ '15.3ms'  │         '87.44%'         │
│    2    │ 'postcss-preset-env + @csstools/postcss-minify' │   '8.5ms'    │   '14.4ms'   │   '18.8ms'   │ '137.4ms' │        '116.60%'         │
│    3    │                    'esbuild'                    │   '0.7ms'    │   '1.4ms'    │   '2.0ms'    │ '18.9ms'  │         '84.24%'         │
│    4    │                 'lightningcss'                  │   '3.7ms'    │   '5.9ms'    │   '9.7ms'    │ '20.5ms'  │        '157.08%'         │
└─────────┴─────────────────────────────────────────────────┴──────────────┴──────────────┴──────────────┴───────────┴──────────────────────────┘
```

This does not show the 100x improvement that `lightningcss` claims to provide 🤷 

Part of it is that the 100x improvement only exists for those using CSSNano in its default config without disabling the `O(2^N)` bugged rules.

Other tools exist for PostCSS 😎 

_~~Notice how `@csstools/postcss-minify` is faster than `lightningcss`, it is doing a lot less work, so again apples and oranges.~~_ -> see https://github.com/parcel-bundler/lightningcss/pull/652#issuecomment-1873406450

-------

It might be interesting to setup a separate repository specifically to benchmark CSS tooling?

A shared project would be able to surface and document the strengths of different tools in a positive and educational way.

It would also give feedback to tool creators on area's that could be improved.

-------

This PR is not intended to be merged and can be closed after reading.
I think this is an area that can be improved but this is not my project :)


devongovett:
Make sure you're using a release build of lightningcss (`yarn build-release` rather than `yarn build`). On my machine, I get the following results with your script:

```
devongovett@Devons-MacBook-Air lightningcss % node bench.js preset-env.css
┌─────────┬─────────────────────────────────────────────────┬──────────────┬──────────────┬──────────────┬───────────┬──────────────────────────┐
│ (index) │                      label                      │ percentile50 │ percentile95 │ percentile99 │    max    │ percentageOfOriginalSize │
├─────────┼─────────────────────────────────────────────────┼──────────────┼──────────────┼──────────────┼───────────┼──────────────────────────┤
│    0    │                    'cssnano'                    │   '6.4ms'    │   '15.3ms'   │   '20.6ms'   │ '239.2ms' │         '82.44%'         │
│    1    │           '@csstools/postcss-minify'            │   '0.8ms'    │   '3.1ms'    │   '5.9ms'    │ '14.1ms'  │         '87.44%'         │
│    2    │ 'postcss-preset-env + @csstools/postcss-minify' │   '9.8ms'    │   '17.9ms'   │   '23.5ms'   │ '132.7ms' │        '116.60%'         │
│    3    │                    'esbuild'                    │   '0.7ms'    │   '1.1ms'    │   '1.9ms'    │ '53.2ms'  │         '84.79%'         │
│    4    │                 'lightningcss'                  │   '0.4ms'    │   '0.6ms'    │   '1.0ms'    │ '25.3ms'  │        '157.08%'         │
└─────────┴─────────────────────────────────────────────────┴──────────────┴──────────────┴──────────────┴───────────┴──────────────────────────┘
```

Other notes:

* Disabling mergeLonghands in cssnano isn't really fair since lightningcss also does that.
* Comparing against postcss-minify without preset-env but keeping it enabled for lightningcss isn't really fair either, but prefixing is almost free in lightningcss so there is almost no difference (unlike when you add preset-env)
* Running only once was a deliberate choice. In the real-world users will not run their minifier more than once. Part of the reason lightningcss (and esbuild) are fast is that it is written in a compiled language that doesn't need to be run many times for v8 to optimize them. This leads to faster real-world performance in the common case.
* Bootstrap is a good test to see how a minifier scales to large files. cssnano in particular tends to slow down a lot the larger the file you throw at it, whereas lightningcss slows down much less. I also chose it to see how minifiers removed old browser prefixes that weren't needed anymore. Regardless, that file originally came from [this benchmark](https://github.com/GoalSmashers/css-minification-benchmark) I was also testing against originally.

Glad to see postcss-minify making impressive progress though!

romainmenke:
> Disabling mergeLonghands in cssnano isn't really fair since lightningcss also does that.

That didn't work and I forgot it was there.
I never figured out how CSSNano config works.

So that didn't really affect the outcome either way.

But this is also exactly the point I am trying to make :)
The current benchmark is flawed because it is comparing apples and oranges.
Different tools are compared in very specific configurations against very specific sources.

> Comparing against postcss-minify without preset-env but keeping it enabled for lightningcss isn't really fair either, but prefixing/downleveling is almost free in lightningcss so there is almost no difference (unlike when you add preset-env)

Exactly. Just like it isn't representative to compare to CSSNano alone. CSSNano doesn't do any of the prefixing/downleveling. Including those would make everything a lot slower still.

> Running only once was a deliberate choice. In the real-world users will not run their minifier more than once.

This isn't true during development.

It is common for people to have single config for dev and production and to just minify everything all the time.

But that is also besides the point :)

A single run is just to flaky.
It might be fast because the machine is cold and doing nothing else.
It might be slow because some other process is hogging resources.

> Part of the reason lightningcss (and esbuild) is fast is that it is written in a compiled language that doesn't need to be run many times for v8 to optimize. This leads to faster real-world performance in the common case.

I know, and I am not trying to demonstrate that it isn't fast.

------

I was very careful to use neutral language in my original post because I am completely uninterested in competitive stuff. I don't want to get side tracked with "what is fair".

I only care about improving all tools in all areas.

I really think that `lightningcss` can show that it is a lot faster than other tools in a good benchmark. The current benchmark just isn't that.

romainmenke:
Benchmark result on my machines with the suggested edits (`build-release`)

```
┌─────────┬─────────────────────────────────────────────────┬──────────────┬──────────────┬──────────────┬───────────┬──────────────────────────┐
│ (index) │                      label                      │ percentile50 │ percentile95 │ percentile99 │    max    │ percentageOfOriginalSize │
├─────────┼─────────────────────────────────────────────────┼──────────────┼──────────────┼──────────────┼───────────┼──────────────────────────┤
│    0    │                    'cssnano'                    │   '6.5ms'    │   '14.2ms'   │   '17.7ms'   │ '262.6ms' │         '82.44%'         │
│    1    │           '@csstools/postcss-minify'            │   '0.7ms'    │   '2.0ms'    │   '3.2ms'    │  '6.0ms'  │         '87.44%'         │
│    2    │ 'postcss-preset-env + @csstools/postcss-minify' │   '8.9ms'    │   '14.9ms'   │   '20.1ms'   │ '133.4ms' │        '116.60%'         │
│    3    │                    'esbuild'                    │   '0.6ms'    │   '1.0ms'    │   '1.3ms'    │ '24.5ms'  │         '84.79%'         │
│    4    │                 'lightningcss'                  │   '0.3ms'    │   '0.5ms'    │   '0.7ms'    │  '3.3ms'  │        '157.08%'         │
└─────────┴─────────────────────────────────────────────────┴──────────────┴──────────────┴──────────────┴───────────┴──────────────────────────┘
```

This indeed shows a nice jump in performance for `lightningcss`.
Maybe good to document how to run the current benchmark script to make it easier for others to get the correct results?

devongovett:
> The current benchmark is flawed because it is comparing apples and oranges. Different tools are compared in very specific configurations against very specific sources.

You could add autoprefixer and postcss-preset-env to cssnano, but that will make it even slower. Alternatively you could remove the targets from the lightningcss configuration, but that makes no difference to the performance. Either way, if anything this would make lightningcss look even better since it is the only one of the tools in the current benchmark doing that.

> This isn't true during development.

Who minifies during development? And even if they did it wouldn't be back to back runs, there'd be time in between for CPU caches to get evicted, memory to accumulate, etc. Regardless, the benchmark isn't meant to test development performance, it's showing a one-shot production build.

> A single run is just to flaky.

What I usually do is run the whole `bench.js` script multiple times, rather than looping inside the script. This is more representative of real-world use for a one-shot build. Running the script multiple times accounts for outside variables like other processes on the machine. Esbuild uses a [very similar methodology](https://esbuild.github.io/faq/#benchmark-details) for their benchmark.

Like I said before, there are many ways to measure. I chose to make this benchmark as close to real-world as possible: running a single build just like a regular user would. We could add additional configuration to the other tools to try to match lightningcss's features more closely, but I think that would only result in even worse performance so I was trying to be generous. 😄

romainmenke:
I think you are assuming that I want `cssnano` to appear better than it is compared to `lightningcss` and that I somehow think that the current benchmark isn't doing it justice.

The opposite is true.
The current benchmark is indeed being way too generous towards `cssnano`.

But at the same time the result of `cssnano` is used to say something about JavaScript tooling for CSS in general, which I think is questionable.

I think using the slowest tool with a source that brings out the worst of the performance issues isn't representative of real world usage. And I think that it would be informative for users to have better benchmarks that do show apples to apples as best as possible.

-----

This is a run against a real world project :
- bundle
- transpiling/fallbacks
- minification

```
┌─────────┬─────────────────────────────────────────────────────────────────────────────┬──────────────┬──────────────┬──────────────┬───────────┬──────────────────────────┐
│ (index) │                                    label                                    │ percentile50 │ percentile95 │ percentile99 │    max    │ percentageOfOriginalSize │
├─────────┼─────────────────────────────────────────────────────────────────────────────┼──────────────┼──────────────┼──────────────┼───────────┼──────────────────────────┤
│    0    │         '@csstools/postcss-bundler + postcss-preset-env + cssnano'          │   '89.2ms'   │  '116.1ms'   │  '148.5ms'   │ '418.6ms' │        '2968.05%'        │
│    1    │ '@csstools/postcss-bundler + postcss-preset-env + @csstools/postcss-minify' │   '31.3ms'   │   '43.5ms'   │   '49.8ms'   │ '60.9ms'  │        '3080.63%'        │
│    2    │                                  'esbuild'                                  │   '4.7ms'    │   '5.6ms'    │   '7.4ms'    │ '26.7ms'  │        '2904.29%'        │
│    3    │                               'lightningcss'                                │   '2.8ms'    │   '4.3ms'    │   '8.1ms'    │ '16.4ms'  │        '2870.08%'        │
└─────────┴─────────────────────────────────────────────────────────────────────────────┴──────────────┴──────────────┴──────────────┴───────────┴──────────────────────────┘
```

Here you can see that `cssnano` (`418.6ms`) is just not representative of JavaScript tooling for CSS. It has very serious performance issues that should be addressed in that package. If users care about build performance they should not use `cssnano` or they should file issues/pr's in that package to improve things.

In this case `lightningcss` (`16.4ms`) is about twice as fast as `esbuild` (`26.7ms`) which itself is about twice as fast as a good `PostCSS` (`60.9ms`) setup.

`lightningcss` is still very fast and no one is debating that.

devongovett:
> Here you can see that cssnano (418.6ms) is just not representative of JavaScript tooling for CSS.

cssnano is one of the most popular CSS minifiers with 10+ million downloads per week. csso and clean-css are also popular, and generally [in the same order of magnitude](https://goalsmashers.github.io/css-minification-benchmark/) of performance, though perhaps a little faster (but those results are kinda old at this point so I'm not sure). `postcss-minify` seems very new and didn't exist when the benchmark was written, but I'm glad to see it seems to be faster than others. 😄

romainmenke:
> I wonder how much difference there is between the tools in resolving/parsing/evaluating.

This is a run where each tool is tested in a separated process, started with `child_process` `spawn`.

`dummy` is a single JS file that only does a file read and no processing.

```
;node bench.js css/application.css
┌─────────┬─────────────────────────────────────────────────────────────────────────────┬──────────────┬──────────────┬───────────┐
│ (index) │                                    label                                    │ percentile50 │ percentile90 │    max    │
├─────────┼─────────────────────────────────────────────────────────────────────────────┼──────────────┼──────────────┼───────────┤
│    0    │                                   'dummy'                                   │   '52.6ms'   │   '54.2ms'   │ '80.4ms'  │
│    1    │               'postcss-import + postcss-preset-env + cssnano'               │  '566.2ms'   │  '582.4ms'   │ '756.5ms' │
│    2    │ '@csstools/postcss-bundler + postcss-preset-env + @csstools/postcss-minify' │  '295.4ms'   │  '340.8ms'   │ '411.5ms' │
│    3    │                                  'esbuild'                                  │   '57.5ms'   │   '58.6ms'   │ '105.8ms' │
│    4    │                               'lightningcss'                                │   '50.6ms'   │   '55.1ms'   │ '73.1ms'  │
└─────────┴─────────────────────────────────────────────────────────────────────────────┴──────────────┴──────────────┴───────────┘
;node bench.js bootstrap-4.css    
┌─────────┬─────────────────────────────────────────────────────────────────────────────┬──────────────┬──────────────┬───────────┐
│ (index) │                                    label                                    │ percentile50 │ percentile90 │    max    │
├─────────┼─────────────────────────────────────────────────────────────────────────────┼──────────────┼──────────────┼───────────┤
│    0    │                                   'dummy'                                   │   '51.4ms'   │   '53.0ms'   │ '69.5ms'  │
│    1    │               'postcss-import + postcss-preset-env + cssnano'               │  '667.6ms'   │  '698.3ms'   │ '832.1ms' │
│    2    │ '@csstools/postcss-bundler + postcss-preset-env + @csstools/postcss-minify' │  '308.3ms'   │  '325.5ms'   │ '395.6ms' │
│    3    │                                  'esbuild'                                  │   '64.6ms'   │   '65.7ms'   │ '96.2ms'  │
│    4    │                               'lightningcss'                                │   '51.4ms'   │   '57.9ms'   │ '69.1ms'  │
└─────────┴─────────────────────────────────────────────────────────────────────────────┴──────────────┴──────────────┴───────────┘
;node bench.js preset-env.css     
┌─────────┬─────────────────────────────────────────────────────────────────────────────┬──────────────┬──────────────┬───────────┐
│ (index) │                                    label                                    │ percentile50 │ percentile90 │    max    │
├─────────┼─────────────────────────────────────────────────────────────────────────────┼──────────────┼──────────────┼───────────┤
│    0    │                                   'dummy'                                   │   '52.4ms'   │   '53.9ms'   │ '69.0ms'  │
│    1    │               'postcss-import + postcss-preset-env + cssnano'               │  '406.9ms'   │  '423.7ms'   │ '478.3ms' │
│    2    │ '@csstools/postcss-bundler + postcss-preset-env + @csstools/postcss-minify' │  '245.4ms'   │  '248.1ms'   │ '282.9ms' │
│    3    │                                  'esbuild'                                  │   '58.6ms'   │   '60.9ms'   │ '94.7ms'  │
│    4    │                               'lightningcss'                                │   '54.2ms'   │   '55.7ms'   │ '56.2ms'  │
└─────────┴─────────────────────────────────────────────────────────────────────────────┴──────────────┴──────────────┴───────────┘
```

------

I do think it's important to somehow measure and document the performance characteristics of various tools.

I think that capturing and communicating these nuances better can lead to a healthier ecosystem of tools.

I might revisit this topic somewhere in the future (if time allows it) and think of some way to catalogue and measure similar tools in ways that show various aspects of each.

It's a spectrum of capabilities and results and I think this is worth exploring.

Thanks 🙇

Issue #654, opened 2024-01-09
`@property` registration with universal syntax doesn't support empty `initial-value`


thecrypticace:
The following CSS produces a parse error but should not.

The [CSS spec](https://drafts.css-houdini.org/css-properties-values-api-1/#initial-value-descriptor) originally listed the value type for `initial-value` as `<declaration-value>` which meant a value was required but that was inconsistent with the behavior of actual custom property declarations. It was changed, as of August 2023, to `<declaration-value>?` which permits empty values and spaces — the same as custom properties.

Chrome already handles this appropriately but it appears that Safari has not yet been updated with the spec change. Aside: the JS API can be used to work around it in Safari as it works correctly.

[...]

PR #657, opened 2024-01-10
Allow whitespace (or nothing) in the `initial-value` of `@property` with universal syntax

RobinMalfait:
When declaring a `@property` with `syntax: '*'`, then the `initial-value` can be empty or can containe whitespace. Currently this errors with an `Unexpected end of input`.

This PR fixes that by allowing the whitespace in the `initial-value` spot.

I know it is 100% correct because we _do_ want to allow whitespace but just not print all of it.

- Right now I added it in a `syntax.rs` file and it's not _only_ for the `initial-value` but for everything that uses that parser with `SyntaxString::Universal` but I'm not 100% sure on how to fix this. A new property on the parser? An additional argument with some options?
- We also "abuse" the `EndOfInput` error in case of `initial-value:;`, this definitely feels hacky but I got it to work this way. Catching the `EndOfInput` allows us to convert it to `Whitespce("")` (also hacky).

I started with adding the tests so that we can work from there, then updated the parser and the printer.

Looking forward to feedback and obvious improvements!

Fixes: #654

### DIFF
--- a/src/values/syntax.rs
+++ b/src/values/syntax.rs
@@ -200,14 +200,7 @@ impl<'i> SyntaxString {
   ) -> Result<ParsedComponent<'i>, ParseError<'i, ParserError<'i>>> {
     match self {
       SyntaxString::Universal => Ok(ParsedComponent::Token(crate::properties::custom::Token::from(
-        match input.next_including_whitespace() {
-          Ok(token) => token,
-          Err(BasicParseError {
-            kind: BasicParseErrorKind::EndOfInput,
-            ..
-          }) => &Token::WhiteSpace(""),
-          Err(e) => return Err(e.into()),
-        },
+        input.next()?,
       ))),
       SyntaxString::Components(components) => {
         // Loop through each component, and return the first one that parses successfully.
--- a/src/values/syntax.rs
+++ b/src/values/syntax.rs
@@ -200,7 +200,14 @@ impl<'i> SyntaxString {
   ) -> Result<ParsedComponent<'i>, ParseError<'i, ParserError<'i>>> {
     match self {
       SyntaxString::Universal => Ok(ParsedComponent::Token(crate::properties::custom::Token::from(
-        input.next()?,
+        match input.next_including_whitespace() {
+          Ok(token) => token,
+          Err(BasicParseError {
+            kind: BasicParseErrorKind::EndOfInput,
+            ..
+          }) => &Token::WhiteSpace(""),
+          Err(e) => return Err(e.into()),
+        },
       ))),
       SyntaxString::Components(components) => {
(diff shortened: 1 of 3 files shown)


devongovett (approved):
lgtm, thanks!

src/values/syntax.rs
```diff
@@ -200,7 +200,14 @@ impl<'i> SyntaxString {
...
       SyntaxString::Universal => Ok(ParsedComponent::Token(crate::properties::custom::Token::from(
-        input.next()?,
+        match input.next_including_whitespace() {
+          Ok(token) => token,
+          Err(BasicParseError {
+            kind: BasicParseErrorKind::EndOfInput,
+            ..
+          }) => &Token::WhiteSpace(""),
+          Err(e) => return Err(e.into()),
+        },
```
devongovett, line 210: I guess as you mentioned we might want this function to require a value in other cases, e.g. when parsing custom at rules (which also use this grammar). One idea would be to move this to `property.rs` when parsing the `initial-value` property. There is already some special handling for universal syntax strings there.

https://github.com/parcel-bundler/lightningcss/blob/b543fe7fc96239d35b55d1d9ff5cf1f3dd170a8a/src/rules/property.rs#L78-L85

so basically something like 

```rust
match syntax.parse_value(&mut parser) {
  Ok(v) => Some(v),
  Err(BasicParseError {
     kind: BasicParseErrorKind::EndOfInput,
      ..
  }) => ParsedComponent::Token(Token::WhiteSpace("")),
  Err(e) => return Err(e)
}

devongovett: or maybe better than checking for an error, you could use `parser.is_exhausted()` to tell if it's already at the end

RobinMalfait: Yep, the only problem I see with `parser.is_exchausted()` is that it will be true for both `initial-value:;` and `initial-value: ;` and in the second case we _do_ want to know the space exists or not to be collected.

The nice part I think is that `is_exhausted()` would be files if an actual error apart from EndOfInput occured. Therefore we can match all `Err` safely, I think.

I've updated the code with your suggestions and with the above as well. I think this looks better than the original version I implemented, does this look good to you too now that it is hoisted?

RobinMalfait: Wait hold on. I think both `:;` and `: ;` should be allowed, but maybe they both end up as `: ;`. That makes it even simpler.

Let me update the code for that.

RobinMalfait: Alright updated. Right now both `initial-value:;` and `initial-value: ;` and even `initial-value:         ;` should be parseable. But in all these scenarios we now print it as `initial-value: ;`.

If you don't agree with this change, we should be able to revert the last commit 👍

devongovett (approved):

PR #665, opened 2024-01-19
When rendering declarations of KeyFrames disable minification

DanielJoyce:
so that KeyFrame transformations are not broken

* Added Todo to TransformList::to_css as it needs to be refactored and we need more context.
* 4 Tests are still broken

This is mostly to get the ball rolling and to see if this fixes people's use cases around keyframes.

### DIFF
--- a/src/properties/transform.rs
+++ b/src/properties/transform.rs
@@ -57,6 +57,10 @@ impl ToCss for TransformList {
       return Ok(());
     }
 
+    // TODO: Decouple minification of representation from minifaction of white space.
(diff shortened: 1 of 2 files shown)


DanielJoyce:
Another option is to do only whitespace minification in TransformList::to_css, and don't try and change the types. 

Ideally there should be a optional minification pass added, so to_css only deals with formatting. 

parse -> [ minify ] -> to_css

DanielJoyce:
Another option would be to turn minify into an enum, expressing white_space only minification and white_space + transformation

devongovett:
Closed by #694

PR #672, opened 2024-02-06
Update ahash to fix compilation issue on nightly

Kobzol:
The `ahash` crate has stopped compiling on `nightly` since today (https://github.com/tkaitchuck/aHash/issues/200). This broke many crates, including this one, and in turn also a part of the Rust compiler benchmark suite, which uses this crate as one of its runtime benchmarks :)

A newer version of `ahash` (`0.8.7`) is needed to fix the compilation issue. This PR updates the `ahash` dependency to fix the compilation of this crate on `nightly`:

```bash
$ cargo +nightly build --no-default-features --features grid,nodejs
```

However, it still won't work with default features, since `parcel_sourcemap` would also need to be updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
+checksum = "77c3a9648d43b9cd48db467b3f87fdd6e146bcc88ab0180006cef2179fe11d01"
 dependencies = [
  "cfg-if",
  "getrandom",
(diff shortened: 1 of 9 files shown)


kornelski:
@devongovett It'd be great to merge this. The grass_compiler dependency has upgraded ahash already.

devongovett (approved):

devongovett:
Looks like the bump has broken the wasm build due to `getrandom` not being available. I'll need to look into what changed in more detail.

mischnic:
Here's how I did it for Parcel, I think I also had the exact same error there:

crates/node-bindings/Cargo.toml:
```toml
getrandom = { version = "0.2", features = ["custom"], default-features = false }
```
and then:
```js
  const {instance} = await WebAssembly.instantiateStreaming(fetch(input), {
    env: {
      ...napi,
      __getrandom_custom: (ptr, len) => {
        let buf = env.memory.subarray(ptr, ptr + len);
        crypto.getRandomValues(buf);
      },
    },
  });
```

PR #684, opened 2024-02-27
fix(visitor): check type for the underlying js object

kwonoj:
This PR aims to make a stopgap workaround for https://github.com/napi-rs/napi-rs/issues/1641.

TL:DR, with latest napi `get_named_property`'s Result does not ensure if it's correct type of the value or not. It can be null|undefined, causes a failure later.

PR guards for the custom visitor logic to look up each properties - check its value type, then coerce into actual value. If upstream napi fixes this behavior it can go away, but anyone uses napi version ~today can benefit by this changes still.


devongovett (approved):
thanks!

Issue #702, opened 2024-03-15
Post-napi cargo-auditable panic


toastal:
```
thread 'main' panicked at cargo-auditable/src/collect_audit_data.rs:77:9:
cargo metadata failure: error: Package `lightningcss-napi v0.1.0 (/build/source/napi)` does not have feature `rayon`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

Nix run cargo-auditable & I’m trying to do my job maintaining the package on the latest version bump but I am not a rust user & don’t know what this means.

Shnatsel:
I can no longer reproduce this issue on `cargo auditable` v0.6.4 and `rustc`/`cargo` v1.76.

PR #713, opened 2024-04-06
napi: fix build error in cargo-auditable

JohnRTitor:
As reported https://github.com/parcel-bundler/lightningcss/issues/702
when trying to build via `cargo auditable build --release --lib --bin=lightningcss --features="cli"` build fails with the following error.
```
   Compiling lightningcss v1.0.0-alpha.55 (/home/masum/Dev-Environment/lightningcss)
    Building [=======================> ] 169/171: lightningcss                            thread 'main' panicked at cargo-auditable/src/collect_audit_data.rs:77:9:
cargo metadata failure: error: Package `lightningcss-napi v0.1.0 (/home/masum/Dev-Environment/lightningcss/napi)` does not have feature `rayon`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

This patch fixes the bug and the build succeeds on cargo auditable.

### DIFF
--- a/napi/Cargo.toml
+++ b/napi/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2021"
 [features]
 default = []
 visitor = ["lightningcss/visitor"]
-bundler = ["dep:crossbeam-channel", "dep:rayon"]
+bundler = ["dep:crossbeam-channel", "rayon"]
 
 [dependencies]
 serde = { version = "1.0.123", features = ["derive"] }


JohnRTitor:
Should close https://github.com/parcel-bundler/lightningcss/issues/702

devongovett:
Why would the dep for crossbeam-channel and rayon be different here? Not sure I understand the issue here.

JohnRTitor:
I am guessing `cargo auditable` does not think of "rayon" as a hard dependency thus producing this error, as you can see in the error message.

This behaviour is not observed in `cargo` however. We are currently using a patch on Nixpkgs to let it build by removing `dep`. https://github.com/NixOS/nixpkgs/commit/5af750a43a74327a3f66cc3ff427aa52f6005c32

JohnRTitor:
I am not sure, something must have changed between the bump to 1.24.1 https://github.com/parcel-bundler/lightningcss/compare/v1.24.0...v1.24.1

JohnRTitor:
@devongovett is there a reason to hold off this PR, or is it blocked on something?

devongovett:
Mainly that I don't understand the problem. AFAICT it is correct: https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies Why do you think this is wrong? I don't know what `cargo-auditable` is but perhaps that has a bug?

JohnRTitor:
[cargo-auditable](https://github.com/NixOS/nixpkgs/pull/223320) is a tool to store the exact versions of dependencies in rust binaries and makes it easier/possible to audit the dependency tree of rust binaries. 

It is a [drop-in replacement](https://www.reddit.com/r/rust/comments/zgz96m/cargo_auditable_can_now_be_used_as_a_dropin/) of cargo. Mainly this lets easier identification of vulnerabilities (CVEs). Nixpkgs by default uses cargo auditable to build rust binaries. 

Stricter checks like these are implemented by cargo-auditable. And hence this PR. cargo auditable is compatible with cargo itself, so this change won't affect builds with cargo.

PR #739, opened 2024-05-12
Add granular CSS Modules options

timneutkens:
## Change in this PR

Adds granular options to switch on/off CSS module scoping for CSS features that are prefixed by default in Lightning CSS currently. For example animation, grid, and custom identifiers.

```js
let { code, map, exports } = transform({
  // ...
  cssModules: {
    animation: true,
    grid: true,
    customIdents: true,
  }
});
```

The reason this is useful is that for example Webpack's CSS modules implementation does not do prefixing of `grid`. If you're migrating to Lightning CSS, like we're doing in Next.js / Turbopack it's helpful to disable the grid prefixing feature to ensure people have a smooth experience migrating.

### Last test failure

The PR is mostly done, just one more case with `animation: false` where it somehow uses `CustomIdent` which in turn means it still prefixes.

CleanShot 2024-05-12 at 23 12 33@2x

### PR feedback

I'm not proficient in writing Rust currently, using working on these smaller changes in Turbopack and other deps of Turbopack to learn as well, let me know if anything should be changed 👍

### Improved `assert_eq` failure reports in tests

While investigating mismatches it was quite hard to read the `assert_eq` for these outputs when they mismatch. Added `pretty_assertions` to add much better diffs for these cases:

Before:

CleanShot 2024-05-12 at 23 17 36@2x

After:

CleanShot 2024-05-12 at 23 18 21@2x

### DIFF
--- a/napi/src/lib.rs
+++ b/napi/src/lib.rs
@@ -716,9 +716,9 @@ fn compile<'i>(
                 Default::default()
               },
               dashed_idents: c.dashed_idents.unwrap_or_default(),
-              animation: c.animation.unwrap_or_default(),
-              grid: c.grid.unwrap_or_default(),
-              custom_idents: c.custom_idents.unwrap_or_default(),
+              animation: c.animation.unwrap_or(true),
+              grid: c.grid.unwrap_or(true),
+              custom_idents: c.custom_idents.unwrap_or(true),
             }),
           }
         } else {
--- a/napi/src/lib.rs
+++ b/napi/src/lib.rs
@@ -848,7 +848,6 @@ fn compile_bundle<
             dashed_idents: c.dashed_idents.unwrap_or_default(),
             animation: c.animation.unwrap_or(true),
             grid: c.grid.unwrap_or(true),
-            keyframes: c.keyframes.unwrap_or(true),
             custom_idents: c.custom_idents.unwrap_or(true),
           }),
         }
--- a/napi/src/lib.rs
+++ b/napi/src/lib.rs
@@ -605,6 +605,9 @@ enum CssModulesOption {
 struct CssModulesConfig {
   pattern: Option<String>,
   dashed_idents: Option<bool>,
+  animation: Option<bool>,
+  grid: Option<bool>,
+  custom_idents: Option<bool>,
 }
 
 #[cfg(feature = "bundler")]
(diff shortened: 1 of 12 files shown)


napi/src/lib.rs
```diff
@@ -713,6 +716,9 @@ fn compile<'i>(
                 Default::default()
               },
               dashed_idents: c.dashed_idents.unwrap_or_default(),
+              animation: c.animation.unwrap_or_default(),
+              grid: c.grid.unwrap_or_default(),
+              custom_idents: c.custom_idents.unwrap_or_default(),
```
devongovett, line 721: These should probably default to true like below.

src/properties/animation.rs
```diff
@@ -58,17 +58,22 @@ impl<'i> ToCss for AnimationName<'i> {
   where
     W: std::fmt::Write,
   {
+    let css_module_animation_enabled =
+      dest.css_module.as_mut().map_or(false, |css_module| css_module.config.animation);
```
devongovett, line 62:
```suggestion
      dest.css_module.as_ref().map_or(false, |css_module| css_module.config.animation);
```

src/values/ident.rs
```diff
@@ -85,7 +107,12 @@ impl<'i> ToCss for DashedIdent<'i> {
   where
     W: std::fmt::Write,
   {
-    dest.write_dashed_ident(&self.0, true)
+    let css_module_custom_idents_enabled = dest
+      .css_module
+      .as_mut()
+      .map_or(false, |css_module| css_module.config.custom_idents);
+
+    dest.write_dashed_ident(&self.0, true, css_module_custom_idents_enabled)
```
devongovett, line 115: `write_dashed_ident` already checks the `dashed_idents` option. I'm not sure that the `custom_idents` option also needs to affect dashed idents as well?

src/properties/animation.rs
```diff
@@ -58,17 +58,22 @@ impl<'i> ToCss for AnimationName<'i> {
...
+
     match self {
       AnimationName::None => dest.write_str("none"),
       AnimationName::Ident(s) => {
         if let Some(css_module) = &mut dest.css_module {
           css_module.reference(&s.0, dest.loc.source_index)
```
devongovett, line 68: need the check here too

devongovett (approved):

PR #741, opened 2024-05-14
Running prettier on the website files

timneutkens:
Noticed while adding docs for #739 that these files weren't handled by prettier before so unrelated changes got into the commit to change the docs. This ensures the prettier config changes are landed on the website.

### DIFF
--- a/website/pages/minification.md
+++ b/website/pages/minification.md
@@ -9,11 +9,11 @@
 Lightning CSS can optimize your CSS to make it smaller, which can help improve the loading performance of your website. When using the Lightning CSS API, enable the `minify` option, or when using the CLI, use the `--minify` flag.
 
 ```js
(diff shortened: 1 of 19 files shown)


website/pages/minification.md
```diff
@@ -172,14 +193,18 @@ In addition to removing defaults, Lightning CSS also omits quotes, whitespace, a
...
 
 minifies to:
 
 ```css
-.foo{background-image:url(logo.png);background-position:50%;font-weight:700}
+.foo {
+  background-image: url(logo.png);
+  background-position: 50%;
+  font-weight: 700;
+}
```
devongovett, line 207: Hmm might need to ignore some of these that are showing exact minified output...

timneutkens: Ah haha, yeah that needs an exclude

PR #796, opened 2024-08-14
feat: Implement pure mode lints for CSS Modules

kdy1:
Relevant code: https://github.com/css-modules/postcss-modules-local-by-default/blob/5eeb3053bc40061022ebc6f41cd207a405484de1/src/index.js#L518-L520

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6890,6 +6890,12 @@ mod tests {
       ".foo /deep/ .bar {width: 20px}",
       ParserError::SelectorError(SelectorError::DanglingCombinator),
     );
+
+    error_test_with_options(
+      "div {width: 20px}",
+      ParserError::SelectorError(SelectorError::PureCssModuleClass),
+    );
+
     minify_test_with_options(
       ".foo >>> .bar {width: 20px}",
       ".foo>>>.bar{width:20px}",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,7 +71,6 @@ mod tests {
     test_with_options(source, expected, ParserOptions::default())
   }
 
-  #[track_caller]
   fn test_with_options<'i, 'o>(source: &'i str, expected: &'i str, options: ParserOptions<'o, 'i>) {
     let mut stylesheet = StyleSheet::parse(&source, options).unwrap();
     stylesheet.minify(MinifyOptions::default()).unwrap();
@@ -83,6 +82,7 @@ mod tests {
     minify_test_with_options(source, expected, ParserOptions::default())
   }
 
+  #[track_caller]
   fn minify_test_with_options<'i, 'o>(source: &'i str, expected: &'i str, options: ParserOptions<'o, 'i>) {
     let mut stylesheet = StyleSheet::parse(&source, options.clone()).unwrap();
     stylesheet.minify(MinifyOptions::default()).unwrap();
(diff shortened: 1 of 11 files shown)


timneutkens (changes requested):
Can you add some more tests, the current test only covers the top level but there's other case where it shouldn't error.

Examples:

- `div .my-class` -> should not error
- `.my-class div` -> should not error
- `div.my-class` -> should not error

These are the cases I tried that compile fine with webpack CSS Modules

```css
div.my-class {
  color: red;
}

a .my-class {
  color: red;
}

.my-class a {
  color: red;
}
```

timneutkens (approved):
Looks good to me now, thanks for adding the additional tests 💯

timneutkens (changes requested):
Realized this is not documented yet, can you make sure there's documentation added as well, it's in this repo in the `website` folder 👍

jantimon:
Could we please also add tests for nesting?
Because that's buggy right now in https://github.com/css-modules/postcss-modules-local-by-default

should not error:

```css
div {
  .my-class {
     color: red; 
  }
}
```

```css
div {
  .my-class {
     span {
       color: red; 
     }
  }
}
```

should error:

```css
div {
  .my-class {
     color: red; 
  }
  color: blue
}
```

devongovett:
What should happen for cases like `.my-class:is(a)`? Currently that errors, presumably while parsing the inner selector inside the `:is`. If this selector had been written as `a.my-class` it would not error so seems like it probably shouldn't in the `:is` case also?

kdy1 (commented):
I introduced a new selector state flag and I resolved the test case you mentioned.

src/lib.rs
```diff
@@ -6890,6 +6899,41 @@ mod tests {
...
+      ".my-class a {color: red;}",
+      "._8Z4fiW_my-class a{color:red}",
+      pure_css_module_options.clone(),
+    );
+    minify_test_with_options(
+      ".my-class:is(a) {color: red;}",
+      "._8Z4fiW_my-class:is(a){color:red}",
+      pure_css_module_options.clone(),
+    );
+
```
jantimon, line 6936: currently this test fails:

```suggestion
    minify_test_with_options(
      "div:has(.my-class) {color: red;}",
      "div:has(._8Z4fiW_my-class){color:red}",
      pure_css_module_options.clone(),
    );
```

jantimon, line 6936: could we please copy some tests from postcss-modules-local-by-default?

```suggestion

    minify_test_with_options(
      ".foo { html &:hover { a_value: some-value; } }",
      "._8Z4fiW_foo{html &:hover{a_value:some-value}}",
      pure_css_module_options.clone(),
    );
```

from:

https://github.com/css-modules/postcss-modules-local-by-default/blob/5eeb3053bc40061022ebc6f41cd207a405484de1/test/index.test.js#L853-L858

jantimon, line 6936: when css modules where introduced ~9 years ago there was no native nesting - but that changed: https://caniuse.com/css-nesting

please add this test:

```suggestion
    minify_test_with_options(
      ".foo { span { color: red; } }",
      "._8Z4fiW_foo{& span{color:red}}",
      pure_css_module_options.clone(),
    );
```

✅ this test works with the current code

jantimon, line 6936: when css modules where introduced ~9 years ago there was no native nesting - but that changed: https://caniuse.com/css-nesting

please add this test:

```suggestion
    minify_test_with_options(
      "html { .foo { span { color: red; } } }",
      "html{._8Z4fiW_foo{& span{color:red}}}",
      pure_css_module_options.clone(),
    );
```

❌ this test case fails:

```
thread 'tests::test_selectors' panicked at src/lib.rs:6941:5:
called `Result::unwrap()` on an `Err` value: Error { kind: SelectorError(PureCssModuleClass), loc: Some(ErrorLocation { filename: "", line: 0, column: 6 }) }
```

kdy1:
I think `html { .foo { span { color: red; } } }` is nearly impossible to handle with the current model

kdy1:
We are detecting from selector passer, not AST visitor

jantimon:
I see -  checking only the selector has worked very well for many years  
Unfortunately css nesting changed that..

this test case works in nextjs using postcss but never will in lightningcss
https://stackblitz.com/edit/stackblitz-starters-ywc6pu?file=app%2Fpage.module.css

devongovett:
I think this should not be allowed:

```css
html {
  .foo {
    /* ... */
  }
}
```

because it's easy to start adding declarations to the root `html` rule itself which is not pure.

However the inverse seems ok:

```css
.foo {
  div {
    /* ... */
  }
}
```

devongovett:
Updated to implement this in the transform pass rather than in the selector parser, which enables handling of nesting as shown above.

devongovett (approved):

jantimon:
thanks @kdy1 and @devongovett for all your work!

PR #824, opened 2024-10-07
feat: add path parsing to BasicShape

noahbald:
This PR adds support for parsing and minifying path definitions, such as `clip-path: path('M0 0L10 10')` which would be minified as `clip-path: path('M0 0 10 10')`

See [here](https://github.com/noahbald/oxvg/blob/main/crates/oxvg_optimiser/src/jobs/convert_path_data.rs#L168) for tests, which covers all the path optimization related tests from SVGO

---
To finish for this PR
- [x] Implement serde derives
- [x] Implement Visit derive
- [x] Implement json-schema

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -991,12 +991,14 @@ checksum = "7f222829ae9293e33a9f5e9f440c6760a3d450a64affe1846486b140db81c1f4"
 
 [[package]]
 name = "oxvg_path"
-version = "0.0.1-beta.2"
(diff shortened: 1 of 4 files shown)


noahbald:
Hi, starting this as a draft as this a first-time contribution. Let me know if I've missed anything! I suspect I still need to make some updates to TypeScript?

noahbald:
Closing this so that I'm not using `master` in my fork, will re-raise soon

PR #835, opened 2024-10-15
fix(css-modules): Do not transform the container name in CSS Modules

kdy1:
Related: https://github.com/vercel/next.js/issues/71233

Webpack css-loader does not transpile `@container`


devongovett:
Hashing container names is intentional. I think webpack just never updated to support newer css features. Do the options added in #739 help here, or do we need to add a new one?

kdy1:
I added an option instead 👍

devongovett:
I think this should already be covered by the `custom_idents` option, but I guess we can have a more granular option like animation too.

devongovett (approved):

Issue #805, opened 2024-09-01
[bug]lighting css will shuffle css properties order


sbyps:
![img_v3_02e8_5f9d4970-3d6e-4ccf-8b7f-f25f0f7f5b0g](https://github.com/user-attachments/assets/2d35a1e7-7da0-4a7b-9196-7b652f191428)
![img_v3_02e8_44f16586-43a8-4f71-8975-a1dc5f0c19bg](https://github.com/user-attachments/assets/89c3a419-3ad9-4794-b74f-0d4ba56cfadf)

matthiask:
The same thing happened to us as well. `height:auto` was ordered after `height:calc(...)` after minimization instead of before. (`height:auto` shouldn't have been there in the first place, but the change was definitely unexpected.)

matthiask:
Here's a minimal example showing the breakage:

```javascript
import fs from "node:fs";
import { transform } from 'lightningcss';

let style = `
.navigation {
    height: auto;
    height: calc(100vh - var(--header-height, 0rem));
}
`

let { code, map } = transform({
  filename: 'style.css',
  code: Buffer.from(style),
  minify: true,
  sourceMap: true
});

fs.writeFileSync(1, code);
console.log()
```

See:

```shell
$ node minify.mjs 
.navigation{height:calc(100vh - var(--header-height,0rem));height:auto}
$
```

The `height:auto` now overrides the `calc()` value instead of being overridden by it.

PR #838, opened 2024-10-17
fix: size prop can't order corretly.

GiveMe-A-Name:
## Example
```css
.foo { 
   height: auto; 
   height: calc(100vh - var(--header-height, 0rem)); 
}
```
After lightningcss minify, css code would like this 
```css
.foo { 
   height: calc(100vh - var(--header-height, 0rem)); 
   height: auto; 
}
```
But it's not expect.
we can maintain the last prop, then delete the repeat prop. Like bellow:
```
.foo { 
   height: calc(100vh - var(--header-height, 0rem)); 
}
```

fix: https://github.com/parcel-bundler/lightningcss/issues/805

### DIFF
--- a/src/declaration.rs
+++ b/src/declaration.rs
@@ -495,7 +495,7 @@ pub(crate) struct DeclarationHandler<'i> {
   flex: FlexHandler,
   grid: GridHandler<'i>,
   align: AlignHandler,
-  size: SizeHandler,
+  size: SizeHandler<'i>,
   margin: MarginHandler<'i>,
   padding: PaddingHandler<'i>,
   scroll_margin: ScrollMarginHandler<'i>,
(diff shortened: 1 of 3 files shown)


sbyps:
anyone can review this PR? @devongovett

devongovett:
One problem with this approach is that it assumes that all unparsed properties are valid and supported by all browser targets. Unparsed properties may be things other than values containing variables. Say CSS adds a new unit that lightningcss doesn't support yet. This will result in an unparsed property value, which means we may need to preserve multiple values for unparsed properties.

```css
.foo {
  width: 200px;
  width: 200abc;
  width: some-keyword;
}
```

If `some-keyword` and `200abc` aren't supported by all browsers, they'll fall back to `200px. In this case we need to preserve all property values.

I think we can do a better job of deduplicating values that are exactly the same, or values that contain `var()`, but that is a much bigger task that should apply across all properties. For now, I'm just going to fix the immediate bug which is that we don't flush the current values when we hit an unparsed property.

Issue #837, opened 2024-10-16
Support `@font-feature-values`


yisibl:
```css
@font-feature-values Inter {
    @character-variant {
        cv05: 5;
        cv08: 8;
        cv10: 10;
    }
    @styleset {
        ss01: 1;
        ss08: 8;
    }
}
```

- Spec: https://drafts.csswg.org/css-fonts/#font-feature-values
- MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/@font-feature-value
- Demo: https://x.com/yisibl/status/1846557957896765763

![image](https://github.com/user-attachments/assets/fa06b4a4-47ce-4d6c-92ad-e52cf12323a9)

![image](https://github.com/user-attachments/assets/605f2ebe-4bf5-45b1-b2ba-efe7987b36dc)

yisibl:
Test case:

- https://github.com/clagnut/TODS/blob/e693d52ad411507b960cf01a9734265e3efab102/tods.css#L116-L142 
- https://github.com/Sorixelle/srxl.me/blob/4eb4f4a15cb2d21356df24c096d6a819cfdc1a99/public/fonts/inter/inter.css#L201-L222
- https://github.com/MihailJP/Inconsolata-LGC/blob/7c53cf455787096c93d82d9a51018f12ec39a6e9/Inconsolata-LGC.css#L65-L91

yisibl:
Note that when compressing, we also need to handle special versions of Firefox that don't remove the ending semicolon.

```css
@font-feature-values FontName {
  font-display: whatever;
}
```
https://x.com/pjchrobak/status/1059229076987301888

mcabansag:
Perhaps this solution will help you figure out the problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

PR #840, opened 2024-10-18
feat: @font-feature-values

CGQAQ:
Closes: #837

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13732,11 +13732,38 @@ mod tests {
       r#"
       @font-feature-values "Fancy Font Name" {
         @styleset { cursive: 1; swoopy: 7 16; }
-        font-display: swap;
         @character-variant { ampersand: 1; capital-q: 2; }
       }
       "#,
-      r#"@font-feature-values Fancy Font Name{font-display:swap;@styleset{cursive:1;swoopy:7 16}@character-variant{ampersand:1;capital-q:2}}"#,
+      r#"@font-feature-values Fancy Font Name{@styleset{cursive:1;swoopy:7 16}@character-variant{ampersand:1;capital-q:2}}"#,
+    );
+    minify_test(
+      r#"
+      @font-feature-values foo {
+          @swash { pretty: 0; pretty: 1; cool: 2; }
+      }
+      "#,
+      "@font-feature-values foo{@swash{pretty:1;cool:2}}",
+    );
+    minify_test(
+      r#"
+      @font-feature-values foo {
+          @swash { pretty: 1; }
+          @swash { cool: 2; }
+      }
+      "#,
+      "@font-feature-values foo{@swash{pretty:1;cool:2}}",
+    );
+    minify_test(
+      r#"
+      @font-feature-values foo {
+          @swash { pretty: 1; }
+      }
+      @font-feature-values foo {
+          @swash { cool: 2; }
+      }
+      "#,
+      "@font-feature-values foo{@swash{pretty:1;cool:2}}",
     );
   }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13662,7 +13662,7 @@ mod tests {
   @annotation { circled: 1; boxed: 2; }
 }
     "#,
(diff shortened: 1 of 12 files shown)


src/parser.rs
```diff
@@ -570,6 +571,15 @@ impl<'a, 'o, 'b, 'i, T: crate::traits::AtRuleParser<'i>> AtRuleParser<'i> for Ne
       //     let family_names = parse_family_name_list(self.context, input)?;
       //     Ok(AtRuleType::WithBlock(AtRuleBlockPrelude::FontFeatureValues(family_names)))
```
CGQAQ, line 572: What is this about? I didn't see there's a `gecko` feature in `Cargo.toml`, so I did not use it

src/rules/font_feature_values.rs
```diff
@@ -0,0 +1,128 @@
...
+#[cfg_attr(feature = "visitor", derive(Visit))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "into_owned", derive(static_self::IntoOwned))]
+pub struct FontFeatureValuesRule<'i, R = DefaultAtRule> {
+  /// The name of the font feature values.
+  pub name: Vec<FontFeatureValuesRuleName<'i>>,
+  /// The declarations within the `@font-feature-values` rule.
+  #[cfg_attr(feature = "serde", serde(borrow))]
+  pub declarations: DeclarationBlock<'i>,
```
devongovett, line 28: As far as I can tell, the only declaration supported in the `@font-feature-values` rule is `font-display`. This should ideally work more like `@font-face` and `@font-palette-values`, where there is a special property type for the values of this at rule rather than the generic DeclarationBlock which applies only to style rules.

src/rules/font_feature_values.rs
```diff
@@ -0,0 +1,128 @@
...
+  /// The rules within the `@font-feature-values` rule.
+  #[cfg_attr(feature = "serde", serde(default = "default_rule_list::<R>"))]
+  pub rules: CssRuleList<'i, R>,
```
devongovett, line 31: These rules are again more specific to `@font-feature-values` than any general at rule. There should be a type that's specific to the sub-rules of `@font-feature-values`. The `@page` rule provides an example of this.

src/parser.rs
devongovett: previously copied from Firefox source code

src/lib.rs
```diff
@@ -13374,6 +13374,87 @@ mod tests {
...
+@font-feature-values "Fancy Font Name" {
+  @styleset { cursive: 1; swoopy: 7 16; }
+  @character-variant { ampersand: 1; capital-q: 2; }
+  @stylistic { two-story-g: 1; straight-y: 2; }
+  @swash { swishy: 1; flowing: 2; }
+  @ornaments { clover: 1; fleuron: 2; }
+  @annotation { circled: 1; boxed: 2; }
+}
+    "#,
+      r#"@font-feature-values Fancy\ Font\ Name{@styleset{cursive: 1; swoopy: 7 16;}@character-variant{ampersand: 1; capital-q: 2;}@stylistic{two-story-g: 1; straight-y: 2;}@swash{swishy: 1; flowing: 2;}@ornaments{clover: 1; fleuron: 2;}@annotation{circled: 1; boxed: 2;}}"#,
```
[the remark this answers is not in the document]
devongovett: I think because these sub-at rules are not actually being parsed, but just stored as raw tokens.

src/rules/font_feature_values.rs
yisibl: CSSWG recommends removing font-display from this at-rule. https://github.com/w3c/csswg-drafts/issues/2973#issuecomment-2418922905

devongovett:
Thanks! I refactored this slightly to handle:

* Using a custom type for declarations so they are fully parsed. With `DeclarationBlock` they'd all be parsed as unknown properties, and things like `!important` shouldn't be allowed.
* Removed the `font-display` since it is not supported by any browser and will be removed from the spec.
* Refactored FontFamily to separate FamilyName from generic family parsing. Generic families shouldn't be allowed here.
* Handled merging duplicate declarations and rules.

PR #878, opened 2024-12-26
Skip generating unnecessary `@supports`

### LINKED ISSUE #711: Transpilation generates unnecessary fallback inside `@supports`
Transpilation generates unnecessary fallback inside `@support`. For example, when `lab()` is used inside `@supports (color: lab(0% 0 0))`, lightningcss generates `@supports (color: lab(0% 0 0))` inside it.

[playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Afalse%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22safari%22%3A917504%7D%2C%22include%22%3A1048575%2C%22exclude%22%3A0%2C%22source%22%3A%22.x%20%7B%5Cn%20%20--x%3A%20%2381828d%3B%5Cn%7D%5Cn%5Cn%40supports%20(color%3A%20lab(0%25%200%200))%20%7B%5Cn%20%20.x%20%7B%5Cn%20%20%20%20--x%3A%20lab(54.6419%25%20.899702%20-5.96)%3B%5Cn%20%20%7D%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)

These kinds of input happens in Vite because Vite runs lightningcss two times. First to transpile CSS to make other plugins able to process CSS, second to minify the final CSS.
Original issue: https://github.com/vitejs/vite/issues/16354

sapphi-red:
This PR introduces `TargetsWithSupportsScope` struct that tracks the parent `@supports`s. This struct has `current` field that contains the current target that is based on the `target` option and the parent `@supports`s.

This makes lightningcss to skip generating unnecessary `@supports`.

closes #711


src/targets.rs
```diff
@@ -225,6 +248,43 @@ impl Targets {
   }
 }
 
+#[derive(Debug)]
+pub(crate) struct TargetsWithSupportsScope {
+  input_targets: Targets,
+  supports_scope_features: IndexSet<Features>,
```
devongovett, line 254: You could probably get away with just a Vec here. Since `Features` is a bit set, it should only accumulate new items the deeper you go. So you could track each `@supports` you enter in a stack, and modify `current` in place:

```rust
fn enter_supports(&mut self, features: Features) -> bool {
  if features.is_empty() || self.current.exclude.contains(features) {
    // Already excluding all features
    return false;
  }

  self.stack.push(supported);
  self.current.exclude.insert(features);
  true
}

fn exit_supports(&mut self) {
  if let Some(last) = self.stack.pop() {
    self.current.exclude.remove(last);
  }
}
```

Then you wouldn't need to iterate over the stack each time and could avoid the hashing cost of IndexSet.

Do you think this would work or am I missing something?

src/targets.rs
```diff
@@ -225,6 +248,43 @@ impl Targets {
   }
 }
 
+#[derive(Debug)]
+pub(crate) struct TargetsWithSupportsScope {
```
devongovett, line 252: Should this just be part of `Targets`?

src/targets.rs
sapphi-red: That's true. I updated the code to use `Vec`.
I added this line to make cases like what I added as a test case work.
https://github.com/parcel-bundler/lightningcss/pull/878/commits/394930ae65afba70a3bf1212908fd163778e9e40#diff-05702e78430359853e5b38404035fa394540368bee185a2cdc936a62b9cd65b0R271

src/rules/supports.rs
```diff
@@ -149,6 +161,31 @@ impl<'i> SupportsCondition<'i> {
...
+  fn get_supported_features(&self) -> Features {
+    const COLOR_P3_SUPPORTS_CONDITION: &str = ColorFallbackKind::P3.supports_condition_value();
+    const COLOR_LAB_SUPPORTS_CONDITION: &str = ColorFallbackKind::LAB.supports_condition_value();
+    fn get_supported_features_internal(value: &SupportsCondition) -> Option<Features> {
+      match value {
+        SupportsCondition::And(list) => list.iter().map(|c| get_supported_features_internal(c)).try_union_all(),
+        SupportsCondition::Declaration { property_id, value } => match property_id {
+          PropertyId::Color => Some(match value.as_ref() {
+            COLOR_P3_SUPPORTS_CONDITION => Features::P3Colors | Features::ColorFunction,
+            COLOR_LAB_SUPPORTS_CONDITION => Features::LabColors,
```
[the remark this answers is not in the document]
sapphi-red: Yeah, it would be nice to expand this in future.

src/targets.rs
sapphi-red: Do you mean putting the fields in `Targets` and making it stateful?

devongovett:
Sorry for the delay! I implemented improved detection of used features by parsing the value in d398c1b86439dfd243a6c2ba627a8e7981118a44. That also applies the exclusions during printing so that features like `light-dark` that are compiled during printing are excluded when in a `@supports` rule as well.

sapphi-red:
Thanks! That sounds great. It'd be great if I can have your input on https://github.com/parcel-bundler/lightningcss/pull/880, too.

PR #880, opened 2024-12-27
Mark import as external

### LINKED ISSUE #479: Mark import as external
I am trying to mark some `@import` as external, in order to preserve them in the final bundle, but I was unable to find the correct configuration. Is it something that can be achieved today? Otherwise, would you consider to add this feature?

As usual, thanks for your great work.

### LINKED ISSUE #555: Resolver can't import stylesheets from URLs
I'm attempting to import a Google fonts stylesheet with the following
```css
@import url('https://fonts.googleapis.com/css2?family=Lilita+One&display=swap');
```

The bundler throws with
```
SyntaxError: No such file or directory (os error 2)
...
loc: { line: 1, column: 1 },
data: 'ResolverError'
```

I'm using `bundle` to bundle this file. 

Version:
```
"lightningcss": "^1.21.5"
```

sapphi-red:
This PR adds a way to mark an `import` as external with `resolver`.

closes #479
closes #555

### DIFF
--- a/node/test/bundle.test.mjs
+++ b/node/test/bundle.test.mjs
@@ -365,7 +365,7 @@ test('resolve return non-string', async () => {
   }
 
   if (!error) throw new Error(`\`testResolveReturnNonString()\` failed. Expected \`bundleAsync()\` to throw, but it did not.`);
-  assert.equal(error.message, 'expect String, got: Number');
+  assert.equal(error.message, 'expect string or boolean, got: Number');
   assert.equal(error.fileName, 'tests/testdata/foo.css');
   assert.equal(error.loc, {
     line: 1,
@@ -412,6 +412,31 @@ test('should support throwing in visitors', async () => {
   }
 
   assert.equal(error.message, 'Some error');
+});
+
+test('external import', async () => {
+  const { code: buffer } = await bundleAsync(/** @type {import('../index').BundleAsyncOptions} */ ({
+    filename: 'tests/testdata/has_external.css',
+    resolver: {
+      resolve(specifier, originatingFile) {
+        if (specifier === './does_not_exist.css' || specifier.startsWith('https:')) {
+          return true;
+        }
+        return path.resolve(path.dirname(originatingFile), specifier);
+      }
+    }
+  }));
+  const code = buffer.toString('utf-8').trim();
+
+  const expected = `
+@import "https://fonts.googleapis.com/css2?family=Roboto&display=swap";
+@import "./does_not_exist.css";
+
+.b {
+  height: calc(100vh - 64px);
+}
+     `.trim();
+  if (code !== expected) throw new Error(`\`testResolver()\` failed. Expected:\n${expected}\n\nGot:\n${code}`);
 });
 
 test.run();
--- a/src/bundler.rs
+++ b/src/bundler.rs
@@ -79,7 +79,7 @@ enum AtRuleParserValue<'a, T> {
 
 struct BundleStyleSheet<'i, 'o, T> {
   stylesheet: Option<StyleSheet<'i, 'o, T>>,
-  dependencies: Vec<u32>,
+  dependencies: Vec<Dependency>,
   css_modules_deps: Vec<u32>,
   parent_source_index: u32,
   parent_dep_index: u32,
@@ -87,6 +87,27 @@ struct BundleStyleSheet<'i, 'o, T> {
   supports: Option<SupportsCondition<'i>>,
   media: MediaList<'i>,
   loc: Location,
+}
+
+#[derive(Debug, Clone)]
+enum Dependency {
+  File(u32),
+  External(String),
+}
+
+/// The result of [SourceProvider::resolve].
+#[derive(Debug)]
+pub enum ResolveResult {
+  /// A file path.
+  File(PathBuf),
+  /// An external URL.
+  External(String),
+}
+
+impl From<PathBuf> for ResolveResult {
+  fn from(path: PathBuf) -> Self {
+    ResolveResult::File(path)
+  }
 }
 
 /// A trait to provide the contents of files to a Bundler.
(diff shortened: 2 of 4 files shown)


node/test/bundle.test.mjs
```diff
@@ -414,4 +414,29 @@ test('should support throwing in visitors', async () => {
   assert.equal(error.message, 'Some error');
 });
 
+test('external import', async () => {
+  const { code: buffer } = await bundleAsync(/** @type {import('../index').BundleAsyncOptions} */ ({
+    filename: 'tests/testdata/has_external.css',
+    resolver: {
+      resolve(specifier, originatingFile) {
+        if (specifier === './does_not_exist.css' || specifier.startsWith('https:')) {
+          return true;
```
sapphi-red, line 423: For now, I made `resolve` to accept `string | true` to focus on the rough design. I think `string | { specifier: string, external?: boolean }` would be nice here. What do you think?

tests/testdata/has_external.css
```diff
@@ -0,0 +1,3 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto&display=swap');
+@import './does_not_exist.css';
+@import './b.css';
```
sapphi-red, line 3: This case can be bundled without any semantics change, but for the following case, it's difficult to keep the semantics.
```css
@import './b.css'; // bundled
@import "https://fonts.googleapis.com/css2?family=Roboto&display=swap"; // externalized
```
related: https://github.com/postcss/postcss-import/issues/536, https://github.com/parcel-bundler/parcel/issues/5840#issuecomment-778857252, https://github.com/evanw/esbuild/issues/465#issuecomment-787048779

Maybe it's fine to simply change the semantics in this case? (both [postcss-import](https://github.com/postcss/postcss-import/issues/536#:~:text=I%20suggest%20we%20close%20this%20issue%20as%20won%27t%20fix) and [esbuild](https://github.com/evanw/esbuild/pull/1293#:~:text=Note%20that%20while,match%20the%20browser.) seems to do that)

src/bundler.rs
```diff
@@ -102,7 +123,7 @@ pub trait SourceProvider: Send + Sync {
 
   /// Resolves the given import specifier to a file path given the file
   /// which the import originated from.
-  fn resolve(&self, specifier: &str, originating_file: &Path) -> Result<PathBuf, Self::Error>;
+  fn resolve(&self, specifier: &str, originating_file: &Path) -> Result<ResolveResult, Self::Error>;
```
sapphi-red, line 126: I'm not familiar with the rust libraries' semver compat, but I guess changing the return value of a public trait is a breaking change. Is it fine to introduce a breaking change? If not, I'll try to add `resolve_advanced` method with a default implementation so that it won't be a breaking change.
(On the JS side it's not a breaking change)

sapphi-red:
@devongovett Would you take this one a look?

devongovett:
As you mentioned, one issue is that it will change the behavior of the CSS if external and internal imports are mixed.

```css
@import "other.css";
@import "http://example.com/external.css";
@import "another.css";
```

If `other.css` and `another.css` get bundled together, but `external.css` remains as an import the order will change. We could potentially enforce that all external imports must be declared ahead of bundled dependencies maybe. What do you think? Is that too restrictive?

sapphi-red:
I think it's too restrictive to error on it. I assume it won't matter in most cases. But I think it'd be nice to have a warning.

node/test/bundle.test.mjs
devongovett: yeah or maybe just `string | {external: string}`?

src/bundler.rs
devongovett: it's ok, the crate is still under pre-release

devongovett:
Currently the bundler produces invalid css when externals are after bundled imports: `other.css` gets inlined and the `external.css` import is placed after it. However CSS requires that `@import` rules are before all other rules, except `@layer` declarations so this won't work. If we don't want to emit an error, we will have to hoist the external imports to the top of the bundle.

However when layers are involved this gets complicated.

```css
@layer a;
@import 'a.css';
@layer b;
@import 'external.css';
```

Here the user has declared layers a and b before `external.css`. Ideally we would retain this. So somehow we would produce this I think:

```css
@layer a;
@layer b;
@import 'external.css';

/* a.css */
```

But what if `@layer b` was actually declared inside `a.css` instead of inline? Then it gets more complicated and we essentially need to track layers globally across all files and interleave external imports in the right places.

It would be much simpler to emit an error if externals are found after bundled imports. I am tempted to try this and see what issues people run into. We could potentially try implementing that later (or even a simpler approach where all externals are hoisted to the top regardless of layers). But perhaps that should be behind an unsafe flag or something. What do you think?

sapphi-red:
Starting with that sounds good to me.

Maybe we can tell users to add `layer(...)`s to external imports that are after bundled imports. If the external imports has a layer function, we can hoist the `@import` and generate `@layer` before the `@import`s based on the correct order.
```css
/* # input */
/* ## index.css */
@import 'a.css'
@import 'external1.css' layer(b);
@import 'c.css'

/* ## a.css */
@layer a;
.a {}

/* ## c.css */
@layer c;
.c {}

/* ---------------- */
/* # output */
@layer a, b;
@import 'external1.css' layer(b);
.a {}
@layer c;
.c {}
```

Just in case, this example:
> ```css
> @layer a;
> @import 'a.css';
> @layer b;
> @import 'external.css';
> ```

is not a valid code as [`@import` must be consecutive](https://www.w3.org/TR/css-cascade-5/#layering:~:text=However%2C%20%40import%20and%20%40namespace%20rules%20must%20be%20consecutive%2C%20without%20any%20intervening%20rules.%20The%20following%20is%20invalid%3A). A correct input would be like:
```css
/* ## index.css */
@import 'a.css'
@import 'b.css'

/* ## a.css */
@layer a;
@import 'external1.css';

/* ## b.css */
@layer b;
@import 'external2.css';
```
which requires layer "b" to come after `external1.css` but before `external2.css`. I guess it's impossible to keep the semantics unless we generate multiple files, or use the `@import 'data:text/css,...'` trick.

devongovett:
Oh right, I forgot about the consecutive requirement. That makes it slightly simpler.

And good idea about the layers. Unfortunately that affects the behavior with unlayered styles too (e.g. inline styles on the page that we can't see) since unlayered styles always take precedence over layered styles. `a.css` and `b.css` would also need to be in layers. Automatically adding layers wouldn't account for other css on the page like inline style elements that we can't see at build time. But perhaps there is something there we can do.

I'd still like to start with an error and see how common of an issue it is. It has a relatively simple fix (moving the external css above the bundled styles), but if that's common in libraries that users can't edit then perhaps we need to add an option.

sapphi-red:
Thank you for getting this over the line!

Issue #908, opened 2025-02-04
Unable to compose class named `default`


mrginglymus:
input:

```css
.default {
  color: blue;
}

.foo {
  composes: default;
  color: red;
}
```

expected output:

```json
{
  "default": {
    "name": "EgL3uq_default",
    "composes": [],
    "isReferenced": false
  },
  "foo": {
    "name": "EgL3uq_foo",
    "composes": [
      {
        "type": "local",
        "name": "EgL3uq_default"
      }
    ],
    "isReferenced": false
  }
}
```

actual output:

```json
{
  "default": {
    "name": "EgL3uq_default",
    "composes": [],
    "isReferenced": false
  },
  "foo": {
    "name": "EgL3uq_foo",
    "composes": [],
    "isReferenced": false
  }
}
```

[Playground Reproduction](https://lightningcss.dev/playground/#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Afalse%2C%22cssModules%22%3Atrue%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6553600%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.default%20%7B%5Cn%20%20color%3A%20blue%3B%5Cn%7D%5Cn%5Cn.foo%20%7B%5Cn%20%20composes%3A%20default%3B%5Cn%20%20color%3A%20red%3B%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%221.29.1%22%7D)

mrginglymus:
Ah, it appears the problem is that https://github.com/parcel-bundler/lightningcss/blob/fc7f350e5b26188941778514a1f0c0ae48282e72/src/values/ident.rs#L34-L37 disallows idents with these reserved names. However, I'm not sure that that is required by the css modules spec; it's certainly not explicitly mentioned, and the class references aren't really custom idents.

PR #909, opened 2025-02-04
Add failing test for composing default class

mrginglymus:
Test case for #908

I'll do my best to see if I can find where/why this is happening...


src/properties/css_modules.rs
```diff
@@ -78,7 +78,7 @@ impl<'i> Parse<'i> for Composes<'i> {
 fn parse_one_ident<'i, 't>(
   input: &mut Parser<'i, 't>,
 ) -> Result<CustomIdent<'i>, ParseError<'i, ParserError<'i>>> {
-  let name = CustomIdent::parse(input)?;
+  let name = CustomIdent(input.expect_ident()?.into());
```
mrginglymus, line 81: Well this fixes the test case above, but I have no idea what I'm doing...

PR #953, opened 2025-04-04
Replace currentColor with currentcolor (lowercase)

teddybradford:
Replaces `currentColor` with `currentcolor` (lowercase) to match what's defined in [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#currentcolor-color) and [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) (see: https://github.com/mdn/content/pull/16592). Also: https://github.com/tailwindlabs/tailwindcss/pull/17510.


devongovett:
This should not matter. CSS identifiers are case insensitive.

PR #954, opened 2025-04-07
feat(parser): Improve error recovery of parser

kdy1:
I'm trying to get some feedback before adding more test cases for error recovery.

Do you think this change sounds good to you?

# Context

 - https://github.com/swc-project/plugins/pull/430

I'm trying to replace `swc_css` with `lightningcss` in `styled-jsx` (and turbopack), but I found an edge case where some more error recovery is necessary due to the way `styled-jsx` works.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,6 +66,7 @@ mod tests {
   use indoc::indoc;
   use pretty_assertions::assert_eq;
   use std::collections::HashMap;
+  use std::sync::{Arc, RwLock};
 
   fn test(source: &str, expected: &str) {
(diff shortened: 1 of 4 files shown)


chenjiahan:
We have seen similar cases among Rspack users, this PR should help us resolve the issues 👍

kdy1:
@devongovett I added one more test, and it seems like it's already good! Can you take a look?

devongovett (approved):
passing the options through everywhere is a bit annoying but overall seems ok

vpontis:
@devongovett thanks! Could we get another `lightningcss` release?

Issue #942, opened 2025-03-27
Add support for picker pseudo-element


lubomirblazekcz:
Support for customizable select was added recently in stable Chrome https://developer.chrome.com/blog/a-customizable-select

Currently `::picker`, `::picker-icon`, `::checkmark` throws warnings

PR #957, opened 2025-04-09
Add support for `::picker`, `::picker-icon` and `::checkmark`

lucasweng:
Closes https://github.com/parcel-bundler/lightningcss/issues/942

Adds support for
  - https://developer.mozilla.org/en-US/docs/Web/CSS/::picker
  - https://developer.mozilla.org/en-US/docs/Web/CSS/::picker-icon
  - https://developer.mozilla.org/en-US/docs/Web/CSS/::checkmark

### DIFF
--- a/src/selector.rs
+++ b/src/selector.rs
@@ -292,6 +292,10 @@ impl<'a, 'o, 'i> parcel_selectors::parser::Parser<'i> for SelectorParser<'a, 'o,
       "-webkit-scrollbar-corner" => WebKitScrollbar(WebKitScrollbarPseudoElement::Corner),
       "-webkit-resizer" => WebKitScrollbar(WebKitScrollbarPseudoElement::Resizer),
 
+      "picker" => Picker,
+      "picker-icon" => PickerIcon,
+      "checkmark" => Checkmark,
+
       "view-transition" => ViewTransition,
 
       _ => {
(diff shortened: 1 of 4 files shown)


pouyakary:
@lucasweng thank you!

src/selector.rs
```diff
@@ -953,6 +958,17 @@ pub enum PseudoElement<'i> {
     /// A part name selector.
     part: ViewTransitionPartSelector<'i>,
   },
+  /// The [::picker](https://drafts.csswg.org/css-forms-1/#the-picker-pseudo-element) pseudo element.
```
devongovett, line 961: Is this actually supported? The spec seems to suggest that only the functional form is allowed.

lucasweng: Only the functional form works for now, so I’ll remove the non-functional form to avoid confusion. I originally added support for it because the spec suggests it may be supported in the future. 😬
CleanShot 2025-04-22 at 11 33 26@2x

lucasweng: Removed the non-functional form and added tests for `::picker`. Thanks! https://github.com/parcel-bundler/lightningcss/pull/957/commits/afac67a3b3a03d26878d036e51a38d4e2c94072d

devongovett:
It's in an official w3c spec, and one of the editors is from Apple so presumably they will support it in the future? I don't see any harm here.

Issue #965, opened 2025-04-20
Suggestion on Expanding Node.js tests


drwpow:
Hello! I was looking at some of the open issues and would love to start submitting bugfixes. But looking at the test suite for Node.js I think it would help contributors like myself if we expanded the test setup to more easily allow slotting in new test cases.

Of course touching tests can get dicey! And we want to improve coverage, not destabilize. So I wanted to talk through a strategy. Here’s what I’d like to propose, but happy to change any of these points

- **Use Vitest**. uvu was great but Vitest supports full ESM and typechecking and can simulate multiple environments (even browsers now!). As it’s now a standard testrunner in many OSS libraries it would be more familiar to most
- **Use `test.each()`**. In libraries like this where there can be so many layers, testing the entire surface (Node.js + wasm) gives the best “bang for the buck” but only if there are a _lot_ of itty-bitty inputs and outputs.

    Testing individual layers can be easy to grok but it requires duplicating tests for every layer. Mega E2E tests that produce large outputs aren’t great at tiny bugfixes. But testing tiny inputs, tiny outputs of the whole thing is the perfect blend of minimal test maintenance + maximum coverage for micro/behaviors. I’d love to introduce a setup closer to that.

- **Using `just` as a wrapper**. I really like the test setup of Biome.js and Rolldown that do more automated setup of the tests where you can be more confident running tests locally matches CI. This is more of a DX thing, but given other OSS projects seem to have smoothed over a lot of these tools, aligning could yield benefits.

- **Not touching Rust tests*+ (for now). The changes I’d like to propose would be ONLY for Node.js tests. Perhaps we decide to make the Rust tests _easier_ to run for folks that aren’t Rustaceans, but IMO those tests are fine and great as they are.

    Should also note that there are a lot of existing Node.js tests that capture a lot of behavior that should be 100% preserved! And for existing tests of course I’m only suggesting a reorganization but not losing any existing coverage. Only making the test suite simpler for folks to extend.

Do you have any thoughts about this? Supposing this plan sounds good, I was thinking the first PR would just be a simple **reorganization** without actually changing any tests, get everything passing, then go from there. Would love thoughts/pushback on any of this.

Thanks for a great library!

PR #967, opened 2025-04-21
test: Use Vitest as test runner

drwpow:
Draft PR as part of #965, just to show suggested starting point


vitest.workspace.ts
```diff
@@ -0,0 +1,27 @@
...
+export default defineWorkspace([
+  {
+    test: {
+      name: 'Node.js',
+      environment: 'node',
+      restoreMocks: true,
+    },
+  },
+
+  // TODO: enable for browser tests
```
drwpow, line 12: Note: I was able to get the browser tests set up locally, but it (correctly) errs on

```
module "browserslistToTargets.js" declares no export named "browserslistToTargets"
```

as it’s CJS, attempting to run as an ESM import. I don’t want to touch any runtime with a 10' pole, especially not in this PR, so maybe there’s a way we could build this file or something just to get the proper behavior in tests.

node/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "moduleResolution": "node",
     "isolatedModules": true,
     "noEmit": true,
-    "strict": true
+    "strict": true,
+    "skipLibCheck": true
```
drwpow, line 9: Necessary for Vitest, unfortunately 😕. But also, this is fine to add IMO as this PR does add more typechecking to tests.

PR #986, opened 2025-05-20
Add test for line comments after closing parenthesis

vpontis:
Right now this tests fails but I believe it should pass once a bug is fixed with Lightningcss.


vpontis:
Ah this is an issue with my understanding, sorry!

Issue #1056, opened 2025-09-18
lightningcss no longer compiles with serde 1.0.220 or above


colinaaa:
The following error occurs after upgrading to serde 1.0.220+

```
error[E0433]: failed to resolve: could not find `__private` in `serde`
    --> src/properties/mod.rs:992:26
     |
992  |             Content(serde::__private::de::Content<'de>),
     |                            ^^^^^^^^^ could not find `__private` in `serde`
...
1199 | / define_properties! {
1200 | |   "background-color": BackgroundColor(CssColor),
1201 | |   "background-image": BackgroundImage(SmallVec<[Image<'i>; 1]>),
1202 | |   "background-position-x": BackgroundPositionX(SmallVec<[HorizontalPosition; 1]>),
...    |
1615 | |   "color-scheme": ColorScheme(ColorScheme),
1616 | | }
     | |_- in this macro invocation
     |
     = note: this error originates in the macro `define_properties` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Should be related with https://github.com/serde-rs/serde/pull/2608

Similar issue: https://github.com/swc-project/swc/issues/11092

PR #1060, opened 2025-09-24
Move away from private serde APIs

mischnic:
Part of #1056

<details>
<summary> Lifetime problem that I figured out already </summary>

```
error[E0309]: the parameter type `R` may not live long enough
   --> src/rules/mod.rs:436:56
    |
436 | #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize), serde(transparent))]
    |                                                        ^^^^^^^^^^^^^^^^^^ ...so that the type `R` will meet its required lifetime bounds
...
439 | pub struct CssRuleList<'i, R = DefaultAtRule>(
    |                        -- the parameter type `R` must be valid for the lifetime `'i` as defined here...
    |
    = note: this error originates in the derive macro `serde::Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider adding an explicit lifetime bound
    |
439 | pub struct CssRuleList<'i, R: 'i = DefaultAtRule>(
    |                             ++++

For more information about this error, try `rustc --explain E0309`.
error: could not compile `lightningcss` (lib) due to 1 previous error
```

</details>

### DIFF
--- a/src/rules/mod.rs
+++ b/src/rules/mod.rs
@@ -195,7 +195,7 @@ pub enum CssRule<'i, R = DefaultAtRule> {
 // Manually implemented deserialize to reduce binary size.
 #[cfg(feature = "serde")]
 #[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
-impl<'de: 'i, 'i, R: serde::Deserialize<'de> + 'i> serde::Deserialize<'de> for CssRule<'i, R> {
+impl<'i, 'de: 'i, R: serde::Deserialize<'de> + 'i> serde::Deserialize<'de> for CssRule<'i, R> {
   fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
   where
     D: serde::Deserializer<'de>,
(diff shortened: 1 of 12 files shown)


src/rules/mod.rs
```diff
@@ -226,131 +223,130 @@ impl<'i, 'de: 'i, R: serde::Deserialize<'de>> serde::Deserialize<'de> for CssRul
...
-        let mut value: Option<serde::__private::de::Content> = None;
+        let mut value: Option<Self::Value> = None;
         while let Some(key) = map.next_key()? {
           match key {
             Field::Type => {
               rule_type = Some(map.next_value()?);
             }
             Field::Value => {
-              value = Some(map.next_value()?);
+              let rule_type = rule_type.as_ref().ok_or_else(|| serde::de::Error::missing_field("type"))?;
```
devongovett, line 233: Is it safe to assume that the type field always comes before the value field? I think that's why it was buffering before.

devongovett (commented):
I think we may need to use a crate that provides similar functionality to the private types so that we can buffer correctly. e.g. serde_value

devongovett:
See #1062

PR #1063, opened 2025-09-29
Make TypeScript return type smarter for "map"

SuperLlama88888:
If `sourceMap` isn't passed as an option to `transform()`, `bundle()` or `bundleAsync()`, the return `map` will be `void`. If it is passed as true, `map` will be returned as `Uint8Array`. If `sourceMap` is a generic boolean, `map` will be `Uint8Array | void`.

This more accurately describes what actually happens when calling transform() or bundle() and removes the need for an assertion or `as`.


SuperLlama88888:
Any updates on this??

kdy1 (commented):
Not sure if this worths the compile time cost (of tsc)

SuperLlama88888:
Considering the complexity of other TypeScript code already in this project...
https://github.com/parcel-bundler/lightningcss/blob/0c594d45c832e76ed61c85ae52a31802445c66e8/node/index.d.ts#L77-L150

...I don't think that this change will impact the tsc compile time much.

PR #1090, opened 2025-11-24
Allow `@property` to be nested inside at-rules

### LINKED ISSUE #968: Unknown at rule: @property
Just something I came across when using the vite-rolldown integration which switched over to lightningcss from esbuild. I am currently getting warnings from lightningcss for the following css. Haven't deep dived into css land so excuse me when this is already known/not an issue.

```css
@layer base {
  @property --radialprogress {
    syntax: "<percentage>";
    inherits: true;
    initial-value: 0%;
  }
}
```

[playground](https://lightningcss.dev/playground/#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Atrue%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22%40layer%20%7B%5Cn%20%20%40property%20--radialprogress%20%7B%5Cn%20%20%20%20syntax%3A%20%5C%22%3Cpercentage%3E%5C%22%3B%5Cn%20%20%20%20inherits%3A%20true%3B%5Cn%20%20%20%20initial-value%3A%200%25%3B%5Cn%20%20%7D%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)

thecrypticace:
This updates things such that `@property` no longer errors when nested inside an at-rule.

Note: `@property` is *not* allowed inside style rules (i.e. it silently acts as if it does nothing in the browser) so I left the `allowed_in_style_rule` condition the way it was.

Question: `test_with_options` doesn't appear to emit warnings inside style rules — I would've expected it to. Assuming this is incorrect any chance you could provide an idea of where to look? (Since `allowed_in_style_rule` is already set up to return false I don't think it's that).

Happy to tweak things / add tests if need be.

Closes #968

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28742,6 +28742,44 @@ mod tests {
     "#,
       "@property --property-name{syntax:\"<color>\";inherits:true;initial-value:#00f}.foo{color:var(--property-name)}",
     );
+
+    test(
+      r#"
+      @media (width < 800px) {
+        @property --property-name {
+          syntax: '*';
+          inherits: false;
+        }
+      }
+    "#,
+      indoc! {r#"
+        @media (width < 800px) {
+          @property --property-name {
+            syntax: "*";
+            inherits: false
+          }
+        }
+    "#},
+    );
+
+    test(
+      r#"
+      @layer foo {
+        @property --property-name {
+          syntax: '*';
+          inherits: false;
+        }
+      }
+    "#,
+      indoc! {r#"
+        @layer foo {
+          @property --property-name {
+            syntax: "*";
+            inherits: false
+          }
+        }
+    "#},
+    );
   }
 
   #[test]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -28742,6 +28742,44 @@ mod tests {
...
+    "#,
+      indoc! {r#"
+        @layer foo {
+          @property --property-name {
+            syntax: "*";
+            inherits: false
+          }
+        }
+    "#},
+    );
```
yisibl: According to the specification, it is recommended to include other at rules in the test.

e.g. 

- `@supports`
- `@scope`
- `@container`

https://github.com/web-platform-tests/wpt/blob/c18682f41f9cf95a3909d76b21aad274abe8914f/css/css-conditional/container-queries/container-inner-at-rules.html#L49-L55

PR #1108, opened 2025-12-11
Fix casing for camel-cased svg values

### LINKED ISSUE #1106: Add SVG 1.1 variants of presentation attribute values
Hey there, I've noticed the following property are missing some values that are expected for SVG 1.1
- `text-rendering` is missing `"geometricPrecision"`
- `shape-rendering` is missing `"crispEdges"` and `"geometricPrecision"`
- `color-interpolation` is missing `"linearRGB"`
- ~`display` is missing `"marker"` and `"compact"`~
- ~`font-family` won't parse quoted names (e.g. `'Arial Unicode MS'`)~

There may be some more while I'm still investigating

I can make a PR for these if these changes are wanted

noahbald:
Closes #1106 

Turns out these values weren't parsing because they didn't have the casing attribute applied

### DIFF
--- a/src/properties/svg.rs
+++ b/src/properties/svg.rs
@@ -211,6 +211,7 @@ pub enum Marker<'i> {
 
 /// A value for the [color-interpolation](https://www.w3.org/TR/SVG2/painting.html#ColorInterpolation) property.
 #[derive(Debug, Clone, Copy, PartialEq, Parse, ToCss)]
+#[css(case = lower)]
 #[cfg_attr(feature = "visitor", derive(Visit))]
 #[cfg_attr(
   feature = "serde",
@@ -249,6 +250,7 @@ pub enum ColorRendering {
 
 /// A value for the [shape-rendering](https://www.w3.org/TR/SVG2/painting.html#ShapeRendering) property.
 #[derive(Debug, Clone, Copy, PartialEq, Parse, ToCss)]
+#[css(case = lower)]
 #[cfg_attr(feature = "visitor", derive(Visit))]
 #[cfg_attr(
   feature = "serde",
@@ -270,6 +272,7 @@ pub enum ShapeRendering {
 
 /// A value for the [text-rendering](https://www.w3.org/TR/SVG2/painting.html#TextRendering) property.
 #[derive(Debug, Clone, Copy, PartialEq, Parse, ToCss)]
+#[css(case = lower)]
 #[cfg_attr(feature = "visitor", derive(Visit))]
 #[cfg_attr(
   feature = "serde",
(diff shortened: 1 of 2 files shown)


src/properties/svg.rs
```diff
@@ -211,6 +211,7 @@ pub enum Marker<'i> {
 
 /// A value for the [color-interpolation](https://www.w3.org/TR/SVG2/painting.html#ColorInterpolation) property.
 #[derive(Debug, Clone, Copy, PartialEq, Parse, ToCss)]
+#[css(case = lower)]
```
noahbald: question: will this cause the value to output as e.g. `"srgb"` instead of `"sRGB"`? Afaik SVG values shouldn't be case-insensitive

noahbald: [This example](https://developer.mozilla.org/en-US/play?uuid=dc46be17d27fae41fc95835bb7f8b3f261627dc8&state=7ZZRT9swEMe%2FiuVpEpVWpUwUJMj2MIH6sqdN2lNe3OSSeL3alX0ZZYjvvrPbBlIC01BgaNqTfdez7%2B%2B7X069ljUtUZ7K1P%2BoxKUuqP6QyaPpJJOiBl3VxOYJWx8zI0RKmhDiVoiLtVquEIQtBWoDys2cKjQYErDOsSm0qQTVIHKL1o21IXAri4q0NUIROT1vCOKtye21aQGl3yZI967VBWupttY5lKpB2gqL4Z7sitWUHoLoyVt%2BQnCNowD2XNaaE0qRPHTk%2FfTemTk2jx6ZPiHNyf00DorHThxOfp8mTbrl2pQzaeuZOshpE7oO5cnkxri6a7QATFpXS8FR6yo1Ijsahwdv9vox2sV4cnYBsYIqX7Q6U4I17QTE3MdhU1pD41ItNQZXbhunwe38Xv8MFx0et83uQ8pYElytLVCchWPThKkO64BwN%2F55wf4cf%2Fsy%2B8QiezJwJN5G%2FIf%2FVcDf9uyv4d%2FFYviPIDxU4eATfu%2Fa%2BCEYa%2BBPyH4CPpXFokNPV0cvPZ2uhbWDS5eUDiThPaOo4F%2BZiHtdG2Ii9oHAxjMNwxdCZviBc7ckr2TWiIOv32axy6Ne0uQ7mXvP%2FzADcdchpNB%2BherqVMzR5ouzzNzEqO8hiFcGaQm8LZRbyJtf&srcPrefix=%2Fen-US%2Fdocs%2FWeb%2FSVG%2FReference%2FAttribute%2Fcolor-interpolation%2F) from MDN doesn't seem show any difference between casing

Issue #251, opened 2022-08-05
minifier: it is not safe to remove the default value of a function in filter


yisibl:
There are several inconsistencies in Chrome's implementation of defaults in older versions, which can cause filters to behave in opposite ways in older browsers. Perhaps other browsers have similar bugs with inconsistent defaults between old and new versions, I didn't continue searching.

[Playground](https://parcel-css.vercel.app/#%7B%22minify%22%3Afalse%2C%22nesting%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A3407872%7D%2C%22source%22%3A%22.foo%20%7B%5Cn%20%20filter%3A%20brightness(1)%3B%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)

## Input
```css
.foo {
  filter: brightness(1);
}
```

## Output
```css
.foo {
  -webkit-filter: brightness(); /* I'm also concerned that `-webkit-filter` does not support omitting the default values. */
  filter: brightness(); /* In older versions of Chrome, the calculated value is 0 */
}
```

## Chrome bugs
- https://bugs.chromium.org/p/chromium/issues/detail?id=964696#c13
- https://bugs.chromium.org/p/chromium/issues/detail?id=798683#c6

I would recommend turning off arguments omitting of the functions in the `filter` by default, it's not worth sacrificing correctness for those few bytes.

PR #1121, opened 2026-01-07
feat(minify): avoid removing the default value of the `filter` function

yisibl:
There are several inconsistencies in Chrome's implementation of defaults in older versions, which can cause filters to behave in opposite ways in older browsers. Perhaps other browsers have similar bugs with inconsistent defaults between old and new versions, I didn't continue searching.

In older versions of Chrome(49), `-webkit-filter: brightness()` also caused crashes.

## Input

```css
.foo {
  filter: brightness(1);
}
```

## Output
```css
.foo {
  -webkit-filter: brightness(); /* I'm also concerned that `-webkit-filter` does not support omitting the default values. */
  filter: brightness(); /* In older versions of Chrome, the calculated value is 0 */
}
```

- https://bugs.chromium.org/p/chromium/issues/detail?id=964696#c13
- https://bugs.chromium.org/p/chromium/issues/detail?id=798683#c6

I would recommend turning off arguments omitting of the functions in the `filter` by default, it's not worth sacrificing correctness for those few bytes.

Fixed: https://github.com/parcel-bundler/lightningcss/issues/251

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27257,9 +27257,23 @@ mod tests {
       ".foo{filter:url(filters.svg#filter-id)}",
     );
     minify_test(".foo { filter: blur(5px); }", ".foo{filter:blur(5px)}");
-    minify_test(".foo { filter: blur(0px); }", ".foo{filter:blur()}");
+    minify_test(".foo { filter: blur(0px); }", ".foo{filter:blur(0)}");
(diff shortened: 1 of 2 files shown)


devongovett:
CSS correctness is one thing. It is another thing entirely to account for 10 year old browser bugs. There is no source of truth for that. When implementing things we need to rely on the spec being correct, otherwise the job is not possible.

Issue #1096, opened 2025-12-01
Grid template shorthand causes unreachable code with `grid-template-areas`


dlebedynskyi:
## Context 

When using `grid` shorthand property together with `grid-template-areas` we are hitting [`unreachable code`](https://github.com/parcel-bundler/lightningcss/blob/master/src/properties/grid.rs#L830)

## Sample

```css

.grid-broken {
    display: grid;
    /* */
    grid: auto-flow / 1fr 2fr 1fr;
    /*  This will cause an error */
    grid-template-areas: ". content .";
}

.grid-broken-working {
   display: grid;
 /* Not using shorthand syntax for grid works  without the issue */
    grid-auto-flow: row;
   
    grid-template-columns: 1fr 2fr 1fr;
    grid-template-areas: ". content .";
}


```

## Error seen

```
lightningcss/lightningcss/src/properties/grid.rs:830:16:
| internal error: entered unreachable code
| stack backtrace:
```

## To see it live

[...]

Issue #1130, opened 2026-01-12
`grid-template` shorthand parsing error


yisibl:
## input

See [playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Afalse%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.grid%20%7B%5Cn%20%20grid-template-areas%3A%20%5C%22a%20a%20a%5C%22%20%5C%22b%20c%20c%5C%22%3B%5Cn%20%20grid-template-rows%3A%2030px%2060px%20100px%3B%5Cn%20%20grid-template-columns%3A%201fr%201fr%201fr%3B%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)

```css
.grid {
  grid-template-areas: "a a a" "b c c";
  grid-template-rows: 30px 60px 100px;
  grid-template-columns: 1fr 1fr 1fr;
}
```

## output

❌ The `grid-template-rows` is **missing the row with a value of 100px.**

```css
.grid {
  grid-template: "a a a" 30px
                 "b c c" 60px
                 / 1fr 1fr 1fr;
}
```

## expected

```css
.grid {
  grid-template: "a a a" 30px
                 "b c c" 60px
                 ". . ." 100px
                 / 1fr 1fr 1fr;
}
```

sapphi-red:
fixed by #1132

PR #1132, opened 2026-01-14
feat: improve `grid-template-areas` handling and `grid` shorthand

yisibl:
- Fixed `grid-template-areas` causing unreachable errors.
- Add support for grid-auto-flow syntax with grid-template-areas
- Handle the case where areas is set but rows/columns is `None`
- Pad grid template areas with `.` for missing rows
- Add tests for various grid shorthand combinations with areas

Fixed: #1096, #1130


yisibl:
@benface Thank you very much for your support. I currently do not have a suitable international card.

Issue #601, opened 2023-10-11
Minify CSS variable fallbacks


KTibow:
Input:
```css
#foo {
  color: white;
}
#bar {
  color: var(--c, white);
}
```
Current output:
```css
#foo{color:#fff}#bar{color:var(--c,white)}
```
Expected output:
```css
#foo{color:#fff}#bar{color:var(--c,#fff)}
```

PR #1143, opened 2026-01-29
feat: minify colors in var() fallback values 

yisibl:
Fixes: https://github.com/parcel-bundler/lightningcss/issues/601


yisibl:
@devongovett ping~

Issue #1136, opened 2026-01-20
[minify] Do not convert `rotate: 0deg` to `rotate: none`


yisibl:
The specification clearly states that `rotate: 0deg` creates a stacking context and containing block, while `rotate: none` does not.

See Demo: https://codepen.io/yisi/pen/QwEJEmr?editors=0100
<img width="622" height="300" alt="image" src="https://github.com/user-attachments/assets/9a460355-bed9-46dc-8cea-e03e23dc360a" />

> All three properties accept (and default to) the value none, which produces no transform at all. In particular, this value does not trigger the creation of a stacking context or containing block for all descendants, while all other values (including “identity” transforms like translate: 0px) create a stacking context and containing block for all descendants, per usual for transforms.
> https://drafts.csswg.org/css-transforms-2/#valdef-translate-none

## input

```css
.foo {
    rotate: 0deg;
}
```

## output

```css
.foo{rotate:none}
```

## expected

```css
.foo{rotate:0deg}
```

Note: The unit “deg” cannot be omitted.

PR #1147, opened 2026-02-09
feat: improved serialization of the `rotate` property

yisibl:
This PR refactors the serialization of the `rotate` property value according to the CSS Transforms 2 specification.

Previously, `rotate: 0deg` was incorrectly serialized as `rotate: none`. The specification clearly states that `rotate: 0deg` creates a stacking context and containing block, while `rotate: none` does not.

See Demo: https://codepen.io/yisi/pen/QwEJEmr?editors=0100
image

> All three properties accept (and default to) the value none, which produces no transform at all. In particular, this value does not trigger the creation of a stacking context or containing block for all descendants, while all other values (including “identity” transforms like translate: 0px) create a stacking context and containing block for all descendants, per usual for transforms.
> https://drafts.csswg.org/css-transforms-2/#valdef-translate-none

This PR fixes this issue and also provides additional compression benefits.

- `minify_test(".foo { rotate: -0deg }", ".foo{rotate:0deg}");`
- `minify_test(".foo { rotate: z 10deg }", ".foo{rotate:10deg}");`
- `minify_test(".foo { rotate: 0 0 1 10deg }", ".foo{rotate:10deg}");`
- `minify_test(".foo { rotate: 10deg 0 0 -1 }", ".foo{rotate:-10deg}");`

Fixes: https://github.com/parcel-bundler/lightningcss/issues/1136

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8281,13 +8281,13 @@ mod tests {
     minify_test(".foo { rotate: acos(cos(45deg))", ".foo{rotate:45deg}");
     minify_test(".foo { rotate: acos(-1)", ".foo{rotate:180deg}");
     minify_test(".foo { rotate: acos(0)", ".foo{rotate:90deg}");
-    minify_test(".foo { rotate: acos(1)", ".foo{rotate:none}");
+    minify_test(".foo { rotate: acos(1)", ".foo{rotate:0deg}");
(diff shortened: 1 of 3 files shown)


yisibl:
Note: https://drafts.csswg.org/css-transforms-2/#:~:text=There%20is%20no,30deg%20are%20equivalent.

Issue #975, opened 2025-05-07
Automatic `:is()` wrapping of pseudo-elements + nested media query produces invalid CSS


echocrow:
[Lightning CSS Playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6553600%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.foo%3A%3Aafter%2C%5Cn.bar%3A%3Aafter%20%7B%5Cn%20%20%40media%20screen%20%7B%5Cn%20%20%20%20color%3A%20red%3B%5Cn%20%20%7D%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)

Input:
```css
.foo::after, .bar::after {
  @media screen {
    color: red;
  }
}
```
Expected output:
```css
@media screen {
  .foo:after, .bar:after {
    color: red;
  }
}
```
Actual output:
```css
@media screen {
  :is(.foo:after, .bar:after) {
    color: red;
  }
}
```

Unfortunately the wrapping with `:is(...)` here produces invalid CSS, because [:is() cannot contain pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/:is#syntax) ([spec](https://drafts.csswg.org/css-nesting/#nest-selector)).

Oddly enough, this is not an issue when the nesting of media query & selector are swapped. The following input…
```css
@media screen {
  .foo::after, .bar::after {
    color: red;
  }
}
```
…correctly produces the expected output.

Somewhat related issues:
- #352: reported the same issue, but with potentially invalid input CSS. Example given here should be valid: [CodePen](https://codepen.io/EchoCrow/pen/xbbzvgQ)
[...]

PR #1156, opened 2026-02-19
fix: avoid wrapping pseudo-element selectors in :is() during nesting compilation

### LINKED ISSUE #352: Ligthning generate :is with pseudo element
Hi,

Considering the following code:

```
.foo {
  &::before {
    .bar & {
      color: blue;
    }
  }
}
```

ParcelCSS will generate the following CSS:

```
.bar :is(.foo:before){color:#00f}
```

[Playground URL](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22nesting%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22source%22%3A%22.foo%20%7B%5Cn%20%20%26%3A%3Abefore%20%7B%5Cn%20%20%20%20.bar%20%26%20%7B%5Cn%20%20%20%20%20%20color%3A%20blue%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)

However, using a pseudo element selector is not valid inside :is, so it should generate this instead:

```
.bar .foo:before{color:#00f}
```

donlion:
## Summary

When compiling nested selectors like `.parent &` where the parent selector contains a pseudo-element (e.g., `.element::after`), the code was incorrectly wrapping the parent in `:is()`, producing invalid CSS.

**Input:**
```css
.element::after {
  .parent & {
    color: red;
  }
}
```

**Before (incorrect):**
```css
.parent :is(.element:after) {
  color: red;
}
```

**After (correct):**
```css
.parent .element:after {
  color: red;
}
```

## Root Cause

The `is_simple()` function returns `false` for selectors with pseudo-elements because they internally contain a `Combinator::PseudoElement` component. This caused the nesting serialization to fall back to using `:is()`, which cannot contain pseudo-elements per CSS specification.

## Fix

Added a check in `serialize_nesting` to detect when the parent selector contains a pseudo-element. In that case, we serialize the selector directly instead of wrapping it in `:is()`.

## Related Issues

- Fixes #352 (original report from 2022)
- Extends the fix from #975 to handle the `.parent &` case (where `&` is not at the start)

## Test Plan

- Added test case for the specific scenario
- All existing tests pass (`cargo test`)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25260,6 +25260,21 @@ mod tests {
       "#},
     );
 
+    nesting_test(
+      r#"
+      .element::after {
+        .parent & {
+          color: red;
+        }
+      }
+      "#,
+      indoc! {r#"
+      .parent .element:after {
+        color: red;
+      }
+      "#},
+    );
+
     nesting_test_no_targets(
       r#"
         .foo {
--- a/src/selector.rs
+++ b/src/selector.rs
@@ -1717,8 +1717,11 @@ where
     // Otherwise, use an :is() pseudo class.
     // Type selectors are only allowed at the start of a compound selector,
     // so use :is() if that is not the case.
+    // Also, :is() cannot contain pseudo-elements, so we must serialize directly in that case.
     if ctx.selectors.0.len() == 1
-      && (first || (!has_type_selector(&ctx.selectors.0[0]) && is_simple(&ctx.selectors.0[0])))
+      && (first
+        || (!has_type_selector(&ctx.selectors.0[0]) && is_simple(&ctx.selectors.0[0]))
+        || ctx.selectors.0[0].has_pseudo_element())
     {
       serialize_selector(ctx.selectors.0.first().unwrap(), dest, ctx.parent, false)
     } else {


devongovett:
what happens if there are multiple selectors?

```css
.foo, .element::after {
  .parent & {
    color: red;
  }
}
```

doing this without `:is` is really difficult if not impossible in some cases to do correctly.

donlion:
Good point, I’ll see if I can come up with something.
Would it be terrible to compile to this?
```
.parent .foo, .parent .element::after {
}
```

yisibl:
Essentially, the current CSS specification does not permit nesting `&` within pseudo-elements, which still requires addressing the case mentioned in this comment. https://github.com/parcel-bundler/lightningcss/issues/975#issuecomment-2890769773

yisibl:
We need to ensure that the CSS in this WPT test case, after being compiled by lightningcss, still displays entirely in green.
http://wpt.live/css/css-nesting/nesting-basic.html

donlion:
> Essentially, the current CSS specification does not permit nesting `&` within pseudo-elements, which still requires addressing the case mentioned in this comment. [#975 (comment)](https://github.com/parcel-bundler/lightningcss/issues/975#issuecomment-2890769773)

Aha! That's news to me.
I suppose it's a matter of opinion on wether it should work and compile invalid css to valid css.
But I think it’s very valid to decide, that it shouldn't :)

What’s your thoughts?

Issue #1080, opened 2025-10-24
3D Transform Values Get Minified Incorrectly


porada:
Here’s how minification affects 3D transforms:

- `transform: scale3d(1, 1, 1)` → `transform:scale(1)`
- `transform: rotate3d(0, 0, 1, 0deg)` → `transform:rotate(0)`

Although these are visually identical, they create different rendering contexts in browsers. This impacts hardware acceleration and can cause animation glitches on elements using shadows or other GPU-accelerated effects. Safari is especially prone to these issues.

### Expected Behavior

3D transform values should be preserved:

- `scale3d(1, 1, 1)`
- `translate3d(0, 0, 0)`
- `rotate3d(0, 0, 1, 0deg)`

### Workaround

Using a fractional non-zero Z component (e.g., `translate3d(0, 0, 0.001px)`) prevents optimization, but is not ideal.

zalishchuk:
You probably can use `translateZ`, but I also encountered this problem when switching to lightningcss minifier. It's confusing that these behaviors can't be toggled with configuration, as translate3d with zero values impacts browser rendering quite differently from zero-valued translate2d.

```css
transform: translateZ(0);
```

Issue #239, opened 2022-07-22
translate3d after parcel build. it will change to translate


DodoMonster:
As the title says

when I use parcel build ，code will be compressed and  optimize

But something strange and unexpected happened

For example , the css code

`transform: translate3d(0, 0, 0)`  will be optimize and change to `transform: translate(0, 0)`

I think this operation goes against our original intention

can I know why and Is there a better way to do this than setting up --no-optimize

devongovett:
How does this cause an issue for you?

mischnic:
Maybe to force a new compositor layer? That probably doesn't happen with `translate(0, 0)`
https://aerotwist.com/blog/on-translate3d-and-layer-creation-hacks/

stevenpetryk:
MDN mentions that a new stacking context is created whenever one or more of these properties has a value other than `none`:

* transform
* filter
* backdrop-filter
* perspective
* clip-path
* mask / mask-image / mask-border

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context#the_stacking_context

Not sure if browsers follow this rule in practice, or whether Devon wants Lightning CSS's minifier to follow the spec (more correct) vs anticipating browser behavior (more practical).

devongovett:
I noticed that cssnano also does this by default: https://www.npmjs.com/package/postcss-reduce-transforms. These days, there are other ways of forcing a compositor layer, e.g. `will-change: transform` or `backface-visibility`.

> 2d and 3d are completely different.

in what way? are there other observable side effects to converting them?

d4tocchini:
This is a bug. Safari is notorious for relying on translate3d to trigger gpu layers for things `mix-blend-mode`.  `backface-visibility` does not do the job. Sure, there may be other ways to force a gpu layer like `will-change`, but that semantic hack comes with more side effects.

PR #1157, opened 2026-02-21
fix(minify): preserve 3d transform semantics during minification

zalishchuk:
Adjust transform minification so we keep 3D transforms as 3D instead of rewriting them into 2D forms.

The motivation is practical: some of these rewrites are mathematically equivalent, but they are not always behavior-equivalent in browsers (especially around compositing/layer promotion in Safari).

## Output Changes

- `translate3d(..., ..., 0)` is no longer lowered to `translate(...)` / `translateY(...)`.
- `scale3d(..., ..., 1)` is no longer lowered to `scale(...)` / `scaleX(...)` / `scaleY(...)`.
- `scale3d(1,1,z)` now canonicalizes to `scaleZ(z)` (including `scale3d(1,1,1) -> scaleZ(1)`), which keeps transforms in the 3D family.
- `rotateZ(...)` stays `rotateZ(...)` (not `rotate(...)`).
- `rotate3d(0,0,1,a)` now normalizes to `rotateZ(a)` (still 3D).
- `translate3d(0,0,0)` now normalizes to `translateZ(0)` (same family as `translate3d(0,0,z) -> translateZ(z)`).
- Updated docs to reflect the new transform minification behavior.

## Example

```css
/* before -> after */
transform: translate3d(2px,0,0);    /* translate3d(2px,0,0) */
transform: translate3d(0,2px,0);    /* translate3d(0,2px,0) */
transform: translate3d(2px,3px,0);  /* translate3d(2px,3px,0) */
transform: translate3d(0,0,0);      /* translateZ(0) */
transform: translate3d(0,0,2px);    /* translateZ(2px) */

transform: scale3d(2,1,1);          /* scale3d(2,1,1) */
transform: scale3d(1,2,1);          /* scale3d(1,2,1) */
transform: scale3d(2,2,1);          /* scale3d(2,2,1) */
transform: scale3d(1,1,2);          /* scaleZ(2) */
transform: scale3d(1,1,1);          /* scaleZ(1) */

transform: rotateZ(20deg);          /* rotateZ(20deg) */
transform: rotate3d(0,0,1,20deg);   /* rotateZ(20deg) */
```

## Related Issues

- https://github.com/parcel-bundler/lightningcss/issues/1080
- https://github.com/parcel-bundler/lightningcss/issues/239

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12687,19 +12687,23 @@ mod tests {
     );
     minify_test(
       ".foo { transform: translate3d(2px, 0px, 0px)",
-      ".foo{transform:translate(2px)}",
+      ".foo{transform:translate3d(2px,0,0)}",
     );
     minify_test(
       ".foo { transform: translate3d(0px, 2px, 0px)",
-      ".foo{transform:translateY(2px)}",
+      ".foo{transform:translate3d(0,2px,0)}",
     );
     minify_test(
       ".foo { transform: translate3d(0px, 0px, 2px)",
       ".foo{transform:translateZ(2px)}",
     );
     minify_test(
+      ".foo { transform: translate3d(0px, 0px, 0px)",
+      ".foo{transform:translateZ(0)}",
+    );
+    minify_test(
       ".foo { transform: translate3d(2px, 3px, 0px)",
-      ".foo{transform:translate(2px,3px)}",
+      ".foo{transform:translate3d(2px,3px,0)}",
     );
     minify_test(".foo { transform: scale(2, 3)", ".foo{transform:scale(2,3)}");
     minify_test(".foo { transform: scale(10%, 20%)", ".foo{transform:scale(.1,.2)}");
@@ -12710,14 +12714,15 @@ mod tests {
     minify_test(".foo { transform: scaleY(2)", ".foo{transform:scaleY(2)}");
     minify_test(".foo { transform: scaleZ(2)", ".foo{transform:scaleZ(2)}");
     minify_test(".foo { transform: scale3d(2, 3, 4)", ".foo{transform:scale3d(2,3,4)}");
-    minify_test(".foo { transform: scale3d(2, 1, 1)", ".foo{transform:scaleX(2)}");
-    minify_test(".foo { transform: scale3d(1, 2, 1)", ".foo{transform:scaleY(2)}");
+    minify_test(".foo { transform: scale3d(2, 1, 1)", ".foo{transform:scale3d(2,1,1)}");
+    minify_test(".foo { transform: scale3d(1, 2, 1)", ".foo{transform:scale3d(1,2,1)}");
     minify_test(".foo { transform: scale3d(1, 1, 2)", ".foo{transform:scaleZ(2)}");
-    minify_test(".foo { transform: scale3d(2, 2, 1)", ".foo{transform:scale(2)}");
+    minify_test(".foo { transform: scale3d(2, 2, 1)", ".foo{transform:scale3d(2,2,1)}");
+    minify_test(".foo { transform: scale3d(1, 1, 1)", ".foo{transform:scale3d(1,1,1)}");
     minify_test(".foo { transform: rotate(20deg)", ".foo{transform:rotate(20deg)}");
     minify_test(".foo { transform: rotateX(20deg)", ".foo{transform:rotateX(20deg)}");
     minify_test(".foo { transform: rotateY(20deg)", ".foo{transform:rotateY(20deg)}");
-    minify_test(".foo { transform: rotateZ(20deg)", ".foo{transform:rotate(20deg)}");
+    minify_test(".foo { transform: rotateZ(20deg)", ".foo{transform:rotateZ(20deg)}");
     minify_test(".foo { transform: rotate(360deg)", ".foo{transform:rotate(360deg)}");
     minify_test(
       ".foo { transform: rotate3d(2, 3, 4, 20deg)",
@@ -12733,7 +12738,7 @@ mod tests {
     );
     minify_test(
       ".foo { transform: rotate3d(0, 0, 1, 20deg)",
-      ".foo{transform:rotate(20deg)}",
+      ".foo{transform:rotateZ(20deg)}",
     );
     minify_test(".foo { transform: rotate(405deg)}", ".foo{transform:rotate(405deg)}");
     minify_test(".foo { transform: rotateX(405deg)}", ".foo{transform:rotateX(405deg)}");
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -12687,19 +12687,23 @@ mod tests {
...
     minify_test(
       ".foo { transform: translate3d(0px, 0px, 2px)",
       ".foo{transform:translateZ(2px)}",
     );
+    minify_test(
+      ".foo { transform: translate3d(0px, 0px, 0px)",
+      ".foo{transform:translateZ(0)}",
+    );
     minify_test(
       ".foo { transform: translate3d(2px, 3px, 0px)",
```
yisibl, line 12705: I believe the original test case here is missing a closing curly brace `}` at the end.

zalishchuk: Absolutely, this pattern has been part of the surrounding transform minify tests and wasn’t introduced by this PR. Honestly, I'm not quite sure why this pattern exists.

yisibl: I suspect this is a typo.

zalishchuk:
@devongovett Anything to add or change here?

yisibl:
Please double-check whether changing `transform:rotate3d(0,0,1,20deg)` to `rotateZ(20deg)` has any side effects, as this differs from the current PR's output with [esbuild](https://esbuild.github.io/try/#dAAwLjI3LjMALS1taW5pZnkgLS1sb2FkZXI9Y3NzIC0tdGFyZ2V0PWNocm9tZTk5AC5mb28gewogIHRyYW5zZm9ybTogcm90YXRlM2QoMCwwLDEsMjBkZWcpOwp9).

src/lib.rs
```diff
@@ -12710,14 +12714,15 @@ mod tests {
...
+    minify_test(".foo { transform: scale3d(1, 2, 1)", ".foo{transform:scale3d(1,2,1)}");
     minify_test(".foo { transform: scale3d(1, 1, 2)", ".foo{transform:scaleZ(2)}");
-    minify_test(".foo { transform: scale3d(2, 2, 1)", ".foo{transform:scale(2)}");
+    minify_test(".foo { transform: scale3d(2, 2, 1)", ".foo{transform:scale3d(2,2,1)}");
+    minify_test(".foo { transform: scale3d(1, 1, 1)", ".foo{transform:scaleZ(1)}");
     minify_test(".foo { transform: rotate(20deg)", ".foo{transform:rotate(20deg)}");
     minify_test(".foo { transform: rotateX(20deg)", ".foo{transform:rotateX(20deg)}");
     minify_test(".foo { transform: rotateY(20deg)", ".foo{transform:rotateY(20deg)}");
-    minify_test(".foo { transform: rotateZ(20deg)", ".foo{transform:rotate(20deg)}");
+    minify_test(".foo { transform: rotateZ(20deg)", ".foo{transform:rotateZ(20deg)}");
```
yisibl, line 12725: Can you add some tests that combine other functions?
For example: `transform: rotate3d(1, 0, 0, var(--foo))`

zalishchuk:
> Please double-check whether changing `transform:rotate3d(0,0,1,20deg)` to `rotateZ(20deg)` has any side effects, as this differs from the current PR's output with [esbuild](https://esbuild.github.io/try/#dAAwLjI3LjMALS1taW5pZnkgLS1sb2FkZXI9Y3NzIC0tdGFyZ2V0PWNocm9tZTk5AC5mb28gewogIHRyYW5zZm9ybTogcm90YXRlM2QoMCwwLDEsMjBkZWcpOwp9).

I checked this one. Spec-wise, `rotateZ(a)` is defined as equivalent to `rotate3d(0,0,1,a)`, so I don’t see a behavioral side effect from that rewrite. Currently Lightning CSS [already](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22x%20%7B%5Cn%20%20transform%3Arotate3d(0%2C0%2C1%2C20deg)%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D) rewrites `rotate3d(0,0,1,a)` to `rotate(a)`, which is a stronger normalization than this PR. This change would only rewrite it to `rotateZ(a)`, which stays in the 3D rotate family.

src/lib.rs
zalishchuk: I think adding a combined test makes sense, but the `var(--foo)` example would go through the unparsed path rather than the typed rotate3d minification path. I can add a combined typed test here instead, e.g., `rotate3d(...) translate3d(...)`, so we cover the normalization in the same code path.

yisibl:
@zalishchuk I've added some test cases in this PR https://github.com/parcel-bundler/lightningcss/pull/1174. You need to rebase master.

devongovett:
> but they are not always behavior-equivalent in browsers (especially around compositing/layer promotion in Safari).

Could you create a reproduction where something is broken due to this optimization? I'm aware that some sites use 3d transforms as a hack to force GPU compositing in certain browsers, but nothing in the actual CSS spec says this is required. There is a better way of telling the browser to optimize these days: `will-change: transform`. This is specifically designed as a performance hint, unlike 3d transforms.

I don't want to get into the business of accounting for bugs or implementation details of browsers in Lightning CSS. It's not possible to test for that. The CSS spec is the only sane source of truth.

I would not be opposed to adding an option to disable this transform though (like you can with other features like colors and nesting).

zalishchuk:
I get where you’re coming from on spec-first behavior, and I agree with that in general.
But in this case I think `translate3d(x, y, 0)` -> `translate(x, y)` is too aggressive in practice.

A lot of teams still intentionally use 3D transforms as a compositing hint, especially in third-party CSS they don’t control. When Lightning rewrites those to 2D, users can lose behavior and they don’t really have a targeted way to opt out.

In my specific case, this is coming from a third-party CSS import, so I can’t rewrite `translate3d(0, 0, 0)` to `will-change: transform` or `translateZ(0)` myself.

I also checked what other minifiers do right now:
- `esbuild` keeps these cases in 3D form (`translateZ` / `translate3d`)
- `cssnano` does the same for `translate3d(x, y, 0) `cases

So for this specific optimization, preserving 3D form seems to be the safer ecosystem default.

I’m okay with adding a flag if needed, but I do think the current default in Lightning is the risky one here, not preserving the 3D form.

PR #1158, opened 2026-02-22
feat: add `minifyWhitespace` override option

zalishchuk:
Added a new `minifyWhitespace` option (JS/NAPI) and matching CLI flags (`--minify-whitespace` / `--no-minify-whitespace`) so users can control output whitespace/newlines separately from semantic minification.

This is mainly to fix the confusing case where `minify: false` would still "beautify" the output (add spaces/newlines) even if the input was already compact.

Behavior:

- If `minifyWhitespace` is omitted, formatting still follows `minify` like before (old behavior preserved).
- If `minifyWhitespace` is provided, behavior is decoupled:
  - `minify` controls semantic optimizations
  - `minifyWhitespace` controls serializer whitespace/newlines

Examples:
- `minify: false, minifyWhitespace: true` -> no semantic optimizations + compact output
- `minify: true, minifyWhitespace: false` -> semantic optimizations + pretty output

This stays non-breaking for existing JS/CLI users.

### DIFF
--- a/node/test/transform.test.mjs
+++ b/node/test/transform.test.mjs
@@ -2,18 +2,18 @@ import { test } from 'uvu';
 import * as assert from 'uvu/assert';
 import {webcrypto as crypto} from 'node:crypto';
 
-let transform, Features;
+let transform, transformStyleAttribute, Features;
 if (process.env.TEST_WASM === 'node') {
-  ({transform, Features} = await import('../../wasm/wasm-node.mjs'));
+  ({transform, transformStyleAttribute, Features} = await import('../../wasm/wasm-node.mjs'));
 } else if (process.env.TEST_WASM === 'browser') {
   // Define crypto globally for old node.
   // @ts-ignore
   globalThis.crypto ??= crypto;
   let wasm = await import('../../wasm/index.mjs');
   await wasm.default();
-  ({transform, Features} = wasm);
+  ({transform, transformStyleAttribute, Features} = wasm);
 } else {
-  ({transform, Features} = await import('../index.mjs'));
+  ({transform, transformStyleAttribute, Features} = await import('../index.mjs'));
 }
 
 test('can enable non-standard syntax', () => {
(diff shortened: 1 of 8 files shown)


node/test/transform.test.mjs
```diff
@@ -68,4 +68,73 @@ test('can disable prefixing', () => {
...
+test('minifyWhitespace can compact output without semantic minification', () => {
+  let res = transform({
+    filename: 'test.css',
+    code: Buffer.from('.a { color: red; } .a { color: blue; }'),
+    minify: false,
+    minifyWhitespace: true
+  });
+
+  assert.equal(res.code.toString(), '.a{color:red}.a{color:#00f}');
+});
```
yisibl, line 84: Perhaps adding some tests for at-rules and CSS functions would be beneficial.

zalishchuk: Done in [`0686196`](https://github.com/parcel-bundler/lightningcss/pull/1158/commits/0686196ee5825e57c9b6cd1154788d3dc1229a8a)

yisibl: Nice work

devongovett:
I don't think this does what you think it does. The `minify` option in the JS API already controls whether the output is compacted (e.g. printed without whitespace). The `StyleSheet#minify` method in the Rust API is just badly named - it should really be called `transform`. It is not connected to the `minify` JS option at all. Changing the meaning of the `minify` option to mean something else and replacing it with `minifyWhitespace` is a breaking change as well. Basically `minify` is already equivalent to what you mean by `minifyWhitespace`.

> This is mainly to fix the confusing case where minify: false would still "beautify" the output (add spaces/newlines) even if the input was already compact.

This is not really possible. The way lightningcss works is that it parses your code to an AST and then re-prints it. That AST does not contain the original whitespace/comments/etc. So when we re-print it, we can either do so with whitespace or without. There's no way to preserve the whitespace as it was in the input.

Was there a reason you need this? Are you asking for a way to disable certain transforms? That's not what you seem to be describing here but it is what you implemented so I would like to clarify the problem.

Issue #99, opened 2022-03-02
CSS Color Level 5


devongovett:
[v1.5.0](https://github.com/parcel-bundler/parcel-css/releases/tag/v1.5.0) adds support for level 4, let's add support for level 5 as well: https://drafts.csswg.org/css-color-5/

- [ ] `color-mix()`
- [ ] `color-contrast()`
- [ ] Relative color syntax
- [ ] `@color-profile`
- [ ] `device-cmyk`
- [ ] convert colors using ICC profiles??

Not sure how stable the spec is. I think Firefox and Safari already have implementations of some of this under a flag.

svgeesus:
Safari TP has support for `color-mix()`, `color-contrast()` and RCS. Firefox and Chrome are investigating `color-mix()` although one of those implementations is currently sRGB-only at the moment. 

Support for CSS Color 4 and (parts of) CSS Color 5, specifically `color-mix(` and `color-contrast()` is part of [Interop 2022](https://hacks.mozilla.org/2022/03/interop-2022/) (see also [Safari blog post on Interop 2022](https://webkit.org/blog/12288/working-together-on-interop-2022/)) so the latest versions of all major browsers are aiming to support this by the end of this year!

Which will intensify the support need for older browsers.

No browser currently has support for `@color-profile` or `device-cmyk` and those parts of the spec are not part of Interop 2022; but support is showing up in the CSS-to-PDF converters. So I could see down-conversion in parcel-css being useful for people targeting print/ebooks as well as Web.

devongovett:
Implemented the relative color syntax in 44a402da404436fbaed684edfc569e99a4403a0f, passing the tests from WPT. 🥳 Only limitation is that it cannot work with variables due to being a build-time transform.

hare0319:
> Implemented the relative color syntax in [44a402d](https://github.com/parcel-bundler/lightningcss/commit/44a402da404436fbaed684edfc569e99a4403a0f), passing the tests from WPT. 🥳 Only limitation is that it cannot work with variables due to being a build-time transform.

Em... Is there a way to perform this, now? Thanks

LinusU:
I've submitted #808 for the `contrast-color` from CSS Color Level 6, which is the evolution of `color-contrast()` from Level 5 mentioned in the op here 🚀

PR #1160, opened 2026-02-25
feat: implementation of `contrast-color()` from the CSS Color 5

yisibl:
The former name of `contrast-color()` was `color-contrast()`. In CSS Color 5, its syntax has been simplified to:
`contrast-color() = contrast-color( <color> )`

Spec: https://drafts.csswg.org/css-color-5/#contrast-color

This function is part of interop-2026 and is already supported in [Firefox 146](https://bugzilla.mozilla.org/show_bug.cgi?id=1682439)、[Safari 26](https://developer.apple.com/documentation/safari-release-notes/safari-26-release-notes#:~:text=Added%20support%20for%20contrast%2Dcolor) and [Chrome 147.0.7707.0](https://chromiumdash.appspot.com/commit/24d3ad2604a0090584257661cd643df47a5b4a98)(experimental)

This implementation solely parses functions, enabling color values within `contrast-color()` to be compressed or simplified.

Part of #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ An extremely fast CSS parser, transformer, and minifier written in Rust. Use it 
   - Logical properties
   * [Color Level 5](https://drafts.csswg.org/css-color-5/)
     - `color-mix()` function
+    - `light-dark()` function
+    - `contrast-color()` function
     - Relative color syntax, e.g. `lab(from purple calc(l * .8) a b)`
   - [Color Level 4](https://drafts.csswg.org/css-color-4/)
     - `lab()`, `lch()`, `oklab()`, and `oklch()` colors
--- a/src/bundler.rs
+++ b/src/bundler.rs
@@ -864,6 +864,9 @@ fn visit_vars<'a, 'b>(
             stack.push(light.0.iter_mut());
             stack.push(dark.0.iter_mut());
           }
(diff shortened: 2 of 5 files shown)


yisibl:
@devongovett PTAL

devongovett:
I think `contrast-color` with a known value should compute to either `black` or `white` at build time, similar to how we calculate the value of relative colors and `color-mix`. There was a previous PR that implemented this (#808), but it also supported parts of the level 6 spec which is not yet final. Perhaps we can scope that down a bit?

yisibl:
This can be done in the next PR. For this PR, I want to focus on implementing the syntax-level parsing.

However, I believe the algorithm currently defined for `contrast-color` in L5 is subpar. It might not be worth implementing this algorithm; it's better to leave it to the browser to handle.

See https://webkit.org/blog/16929/contrast-color/#:~:text=you%E2%80%99ll%20find%20many%20of%20the%20pairings%20with%20mid%2Dtone%20background%20colors%20don%E2%80%99t%20result%20in%20enough%20contrast.

devongovett:
> This can be done in the next PR.

It would change what gets stored in the AST though right? We wouldn't need a `CssColor::ContrastColor` variant if we parsed it directly to a color.

PR #1170, opened 2026-03-02
Allow visitors to add dependencies

### LINKED ISSUE #877: Feature Request: add a way to register dependencies by plugins
Plugins may rely on a different file but there's no way to tell that information to lightningcss.

For example, the plugin below depends on `.json` placed next to the input file.
```js
export function fooVisitor() {
  /** @type {string[]} */
  let currentStyleSheetSources
  return {
    StyleSheet(stylesheet) {
      currentStyleSheetSources = stylesheet.sources
    },
    Rule: {
      unknown: {
        foo(rule) {
          const from = currentStyleSheetSources[rule.loc.source_index]
          const importedFile = JSON.parse(fs.readFileSync(from + '.json', 'utf8'))
          const cssAst = convertJsonToAst(importedFile) // convert json to lightningcss somehow
          return cssAst
        },
      },
    },
  }
}
```
To make the re-transform happen when the json file is changed, the user that calls `lightningcss.bundle` needs to know which css file depends on which json file.

Simply having a convention (e.g. expose `dependencies` property from the visitor) would work if `composeVisitors` supports it, but having the information in `TransformResult.dependency` feels natural.

devongovett:
Closes #877

Visitors can now be functions that receive an `addDependency` function. When called, this adds either a file or glob dependency that is returned to the caller (e.g. a bundler). Lightning CSS doesn't do anything with these dependencies by default: the caller is responsible for implementing file watching or cache invalidation based on these dependencies. `composeVisitors` also has support for visitors that are functions, passing the options to each visitor.

This is implemented entirely in the JavaScript layer. Dependencies do not cross the Rust boundary. They are appended to the list of other dependencies such as imports and urls returned by Rust when `analyzeDependencies` is true.

### DIFF
--- a/node/composeVisitors.js
+++ b/node/composeVisitors.js
@@ -1,14 +1,22 @@
 // @ts-check
 /** @typedef {import('./index').Visitor} Visitor */
+/** @typedef {import('./index').VisitorFunction} VisitorFunction */
 
 /**
  * Composes multiple visitor objects into a single one.
- * @param {Visitor[]} visitors 
- * @return {Visitor}
+ * @param {(Visitor | VisitorFunction)[]} visitors 
+ * @return {Visitor | VisitorFunction}
  */
 function composeVisitors(visitors) {
   if (visitors.length === 1) {
     return visitors[0];
+  }
+  
+  if (visitors.some(v => typeof v === 'function')) {
+    return (opts) => {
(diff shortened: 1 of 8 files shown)


sapphi-red (approved):
Thank you! I've tested this PR at https://github.com/vitejs/vite/pull/21748 and it worked for my use case

devongovett:
Great. FYI it's released in v1.32.0.

Issue #921, opened 2025-02-21
@property initial-value parsing error


barvian:
[Playground repro](https://lightningcss.dev/playground/#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22%40property%20--my-color%20%7B%5Cn%20%20syntax%3A%20%5C%22%3Ccolor%3E%5C%22%3B%5Cn%20%20inherits%3A%20true%3B%5Cn%20%20initial-value%3A%20color-mix(in%20srgb%2C%20currentColor%2015%25%2C%20transparent)%3B%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)

```css
@property --my-color {
  syntax: "<color>";
  inherits: true;
  initial-value: color-mix(in srgb, currentColor 15%, transparent);
}
```

It seems this may be against the spec, but it works in Chrome/Firefox/Safari and I'm not sure of an alternative 🤔.

devongovett:
Thanks, this is a bug. Currently `color-mix` is only parsed if we can resolve it to a value at build time, and `currentColor` prevents that. We should preserve the original value as is in that case.

PR #1173, opened 2026-03-05
test: remove `unwrap()` in test helpers and print explicit error line context

yisibl:
Previously, adding the example from #921 to the tests caused `unwrap()` issues. This PR replace `unwrap()` based test helper failures with explicit error handling so failures show actionable diagnostics instead of generic panic messages.

- Extract shared test error formatting into `src/test_helpers.rs`
- Print stage-specific failures (`parse/minify/print`)
- Include caller file line with `#[track_caller]`
- Render CSS error line/column with nearby source lines and caret indicator

Part of #921

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,9 @@ pub mod visitor;
 mod serialization;
 
 #[cfg(test)]
+mod test_helpers;
+
+#[cfg(test)]
 mod tests {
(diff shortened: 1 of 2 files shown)


yisibl:
@devongovett PTAL

devongovett (approved):
cool! thanks

PR #1174, opened 2026-03-06
Convert the percentage in the `scale` property or `scale()` to a number

yisibl:
- Use shared scale component parsing so individual `scale` and transform scale functions normalize percentages into numbers consistently.
- Keep serialization canonical by converting scale values through numeric output paths, including `scaleX/Y/Z`.
- Expand transform tests to mirror the individual `scale` percentage cases one-for-one, and add direct AST serialization coverage for both `scale` and `transform`.


yisibl:
@devongovett PTAL

devongovett (approved):

Issue #1184, opened 2026-03-16
Support CSS Overflow Module Level 5


yisibl:
https://drafts.csswg.org/css-overflow-5

- [ ] scroll-marker-group
- [ ] [::scroll-marker-group](https://drafts.csswg.org/css-overflow-5/#selectordef-scroll-marker-group)
- [ ] [::scroll-marker](https://drafts.csswg.org/css-overflow-5/#selectordef-scroll-marker)
- [ ] `::scroll-button( '*' | <scroll-button-direction> )`
- [x] [:target-current](https://drafts.csswg.org/css-overflow-5/#selectordef-target-current), [:target-before](https://drafts.csswg.org/css-overflow-5/#selectordef-target-before) and [:target-after](https://drafts.csswg.org/css-overflow-5/#selectordef-target-after)

## Low priority

- [ ] ::nth-fragment()

PR #1185, opened 2026-03-16
Supports `:target-before/after/current` pseudo-class

### LINKED ISSUE #1181: 'target-current' is not recognized as a valid pseudo-class
Ligntning CSS is throwing a warning

```
[WARN] [vite] [vite:css][lightningcss] 'target-current' is not recognized as a valid pseudo-class. Did you mean '::target-current' (pseudo-element) or is this a typo?
```

for selector:
```css
a:target-current {
  anchor-name: --toc-target;
}
```
[:target-current](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:target-current) is a valid pseudo-class though

yisibl:
Spec: https://drafts.csswg.org/css-overflow-5/#active-before-after-scroll-markers

MDN:

- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:target-current#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:target-after#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:target-before#browser_compatibility

Part of #1184
Fixes: #1181

### DIFF
--- a/c/lightningcss.h
+++ b/c/lightningcss.h
@@ -23,6 +23,7 @@ typedef struct ParseOptions {
   const char *filename;
   bool nesting;
   bool custom_media;
+  bool scroll_navigation_controls;
   bool css_modules;
   const char *css_modules_pattern;
   bool css_modules_dashed_idents;
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -77,6 +77,7 @@ pub struct ParseOptions {
   filename: *const c_char,
   nesting: bool,
   custom_media: bool,
+  scroll_navigation_controls: bool,
   css_modules: bool,
   css_modules_pattern: *const c_char,
   css_modules_dashed_idents: bool,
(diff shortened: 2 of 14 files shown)


devongovett:
Should this be behind a flag or something? My understanding is that this spec is still in early draft/prototype stage, only supported in Chrome, and both [WebKit](https://github.com/WebKit/standards-positions/issues/447) and [Firefox](https://github.com/mozilla/standards-positions/issues/1161) have concerns with it ([W3C TAG too](https://github.com/w3ctag/design-reviews/issues/1037)). Seems early to treat as standard. The current state isn't the worst because it's just a warning, but I suppose we could add a draft flag. wdyt?

yisibl:
I think that's fine.

yisibl:
@devongovett Add `scroll_navigation_controls` flag in https://github.com/parcel-bundler/lightningcss/pull/1185/changes/e200dd5caaf9bb3e110d23bf25bce9d720c7f6e3

devongovett (approved):

PR #1190, opened 2026-03-22
Replace `atty` with `std::io::IsTerminal`

fkeiler:
The [_atty_ repository](https://github.com/softprops/atty) warns that it is no longer maintained and should be replaced with the [equivalent native method](https://doc.rust-lang.org/std/io/trait.IsTerminal.html). This change removes it as a direct dependency (however it stays a transitive dependency through `clap@3`).

`yarn test` and `cargo test` both ran successfully after the change, but the `Not reading from stdin as it was not redirected` error is not checked in any of the test cases, so I also built the CLI with 
```sh
cargo build --bin=lightningcss --features=cli --release
```
and ran `./target/release/lightningcss` on a terminal emulator without any input files nor piping, which generated the expected 
```sh
Error: Custom { kind: Other, error: "Not reading from stdin as it was not redirected" }
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -759,7 +759,6 @@ dependencies = [
  "ahash 0.8.12",
  "assert_cmd",
  "assert_fs",
- "atty",
  "bitflags 2.6.0",
  "browserslist-rs",
  "clap",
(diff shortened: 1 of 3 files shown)


fkeiler:
Wrong branching. Fixed on #1197

PR #1209, opened 2026-04-06
Preserve ordering of `font-*` properties that follow a `font` shorthand

### LINKED ISSUE #1207: Property reordering during minification causes certain settings to be ignored
Lightning's property-reordering behavior during minification can result in shorthand properties being moved *after* a lower-level setting, causing it to be reset to its default value. 

In particular, I've encountered it in this situation ([playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22p%7B%5Cn%20%20font%3A%2012px%20serif%3B%5Cn%20%20font-variant-numeric%3A%20oldstyle-nums%3B%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)):

#### Original CSS
```css
p{
  font: 12px serif;
  font-variant-numeric: oldstyle-nums;
}
```

#### Minified
```css
p{font-variant-numeric:oldstyle-nums;font:12px serif}
```

The `font` shorthand property resets all the `font-variant-*` settings to their default, despite the clear intent of the original CSS to refine font settings *after* selecting the family and size.

## Workaround

Adding `!important` to the `font-variant-numeric` property seems to have the side effect of reordering it to the *end* of the block rather than the beginning ([playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22p%7B%5Cn%20%20font%3A%2012px%20serif%3B%5Cn%20%20font-variant-numeric%3A%20oldstyle-nums%20!important%3B%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)), which solves the immediate rule-override issue but has the knock-on effect of interfering with the normal cascade for the variant setting. So it's not a fully satisfactory ‘solution’ to the problem.

## Proposed Solutions

Reordering should either be something that can be optionally disabled, or the reordering logic should consider which shorthand properties override others and preserve their relative ordering in those cases. Perhaps this latter logic already exists and the `font-variant-numeric` edge case simply needs to be incorporated? 

## See Also

#231 
#276

samizdatco:
This PR prevents non-shorthand `font-*` properties being reordered to precede a shorthand `font` rule. 

## Existing behavior
As described in #1207, the `FontHandler` currently watches for all the properties that could be combined into a shorthand (e.g., `font-family`, `font-weight`, `font-style`, etc.), but will immediately emit any unrecognized font properties such as `font-variant` or `font-kerning` while deferring the `font` shorthand.

For example ([playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.foo%7B%5Cn%20%20font%3A%2012px%20serif%3B%5Cn%20%20font-variant-numeric%3A%20oldstyle-nums%3B%5Cn%20%20font-kerning%3A%20none%3B%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D)):
```css
.foo{
  font: 12px serif;
  font-variant-numeric: oldstyle-nums;
  font-kerning: none;
}
```
will be minified to:
```css
.foo{font-variant-numeric:oldstyle-nums;font-kerning:none;font:12px serif}
```
This is a problem because placing the `font` shorthand at the end has the side-effect of resetting all prior `font-*` properties to their defaults, which is clearly not the intent of the original css.

## Changes

I added a new `match` branch to `FontHandler`’s `handle_property` method to catch any `Property::Custom` that begins with `"font-"` (but has not already been caught as a property that could be merged into the shorthand) and trigger a flush of the pending shorthand *before* emitting the newly encountered longhand.

Initially I applied this to just the `font-variant-*` properties (cd12e58587edc44aa6680de6dc42b9f8e2920b02), but I believe this behavior is actually desirable for *all* non-shorthand font properties so I extended it to the whole `font-*` prefix.

Fixes #1207

## Questions

I'm not quite clear on the distinction between `Property::Custom` and `Property::Unparsed`; maybe this needs to check `Unparsed` properties to see if they match the `font-*` prefix as well?

~I added an optimization that [discards](https://github.com/parcel-bundler/lightningcss/pull/1209/changes#diff-6da1aeda09d0ffb9c25b08c5ede95a89e5367b64490f7374c432fd78886d4182R880) the cached longhand properties when a shorthand `font` declaration is encountered (since they'll all be reset to their defaults anyway). Are there any negative consequences to doing this?~ [*reverted: see discussion below*]

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6653,7 +6653,7 @@ mod tests {
     );
     minify_test(
       ".foo { font-variant-numeric: oldstyle-nums; font: 10px serif; }",
-      ".foo{font:10px serif}",
+      ".foo{font-variant-numeric:oldstyle-nums;font:10px serif}",
     );
     minify_test(
       ".foo { font: 10px serif; font-variant-numeric: oldstyle-nums; font-variant-alternates: stylistic(my-style); }",
@@ -6665,7 +6665,7 @@ mod tests {
     );
     minify_test(
       ".foo { font-kerning: none; font: 10px serif; }",
-      ".foo{font:10px serif}",
+      ".foo{font-kerning:none;font:10px serif}",
     );
     minify_test(
       ".foo { font: 10px serif; font-kerning: none; }",
@@ -6673,7 +6673,7 @@ mod tests {
     );
     minify_test(
       ".foo { font-optical-sizing: none; font: 10px serif; }",
-      ".foo{font:10px serif}",
+      ".foo{font-optical-sizing:none;font:10px serif}",
     );
     minify_test(
       ".foo { font: 10px serif; font-optical-sizing: none; }",
@@ -6681,19 +6681,11 @@ mod tests {
     );
     minify_test(
       ".foo { font-synthesis-style: oblique-only; font: 10px serif; }",
-      ".foo{font:10px serif}",
+      ".foo{font-synthesis-style:oblique-only;font:10px serif}",
     );
     minify_test(
       ".foo { font: 10px serif; font-synthesis-style: oblique-only; }",
       ".foo{font:10px serif;font-synthesis-style:oblique-only}",
-    );
-    minify_test(
-      ".foo { font-kerning: none; }",
-      ".foo{font-kerning:none}",
-    );
-    minify_test(
-      ".foo { font-kerning: none; font: 10px serif; font-optical-sizing: none; }",
-      ".foo{font:10px serif;font-optical-sizing:none}",
     );
   }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6663,6 +6663,30 @@ mod tests {
       ".foo { font: 10px serif; font-variant: small-caps; }",
       ".foo{font:10px serif;font-variant:small-caps}",
     );
+    minify_test(
+      ".foo { font-kerning: none; font: 10px serif; }",
+      ".foo{font-kerning:none;font:10px serif}",
+    );
+    minify_test(
+      ".foo { font: 10px serif; font-kerning: none; }",
+      ".foo{font:10px serif;font-kerning:none}",
+    );
+    minify_test(
+      ".foo { font-optical-sizing: none; font: 10px serif; }",
+      ".foo{font-optical-sizing:none;font:10px serif}",
+    );
+    minify_test(
+      ".foo { font: 10px serif; font-optical-sizing: none; }",
+      ".foo{font:10px serif;font-optical-sizing:none}",
+    );
+    minify_test(
+      ".foo { font-synthesis-style: oblique-only; font: 10px serif; }",
+      ".foo{font-synthesis-style:oblique-only;font:10px serif}",
+    );
+    minify_test(
+      ".foo { font: 10px serif; font-synthesis-style: oblique-only; }",
+      ".foo{font:10px serif;font-synthesis-style:oblique-only}",
+    );
   }
 
   #[test]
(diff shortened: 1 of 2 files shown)


stof:
> I added an optimization that [discards](https://github.com/parcel-bundler/lightningcss/pull/1209/changes#diff-6da1aeda09d0ffb9c25b08c5ede95a89e5367b64490f7374c432fd78886d4182R880) the cached longhand properties when a shorthand `font` declaration is encountered (since they'll all be reset to their defaults anyway). Are there any negative consequences to doing this?

the downside is that this assumes that the shorthand will be parsed successfully by all browsers.
If the shorthand uses a syntax that is not supported by the browser, it would be discarded and would *not* reset all the longhands.

samizdatco:
> the downside is that this assumes that the shorthand will be parsed successfully by all browsers. If the shorthand uses a syntax that is not supported by the browser, it would be discarded and would _not_ reset all the longhands.

Good point. So in that case, maybe the better solution is the one in the [previous commit](https://github.com/parcel-bundler/lightningcss/pull/1209/commits/be85c3f274a59292e34576e0667822cf365a90ed) where the appearance of a longhand property triggers any pending shorthand to be flushed first (i.e., it prevents reordering but doesn't discard potentially-reset longhands)?

Issue #1163, opened 2026-02-26
Add support to ppc64/s390 and IBM AIX architectures


ovflowd:
Hey there 👋, we've been trying to use lightningcss from within nodejs/node, but unfortunately we're stuck with the wasm bindings at the moment as lightningcss doesn't seem to provide prebuilt binaries for the ppc64/s390 and the aix architectures. I wanted to ask if these could be supported/or having binaries being compiled for them?

Example failure: https://ci.nodejs.org/job/node-test-commit-linuxone/nodes=rhel9-s390x/53681/console

IronBlood:
It looks like the latest jobs of `node-test-commit-linuxone` ran successfully?

Anyway here are the [workflow file](https://github.com/IronBlood/build-lightningcss/blob/9bbcd324e5447db7fae4bea19a1ed8f89d164036/.github/workflows/upstream-ppc-riscv.yml) and test results for [s390](https://github.com/IronBlood/build-lightningcss/actions/runs/23648625029/job/68887930522) and [ppc64le](https://github.com/IronBlood/build-lightningcss/actions/runs/23648625029/job/68887930506) if anyone need it or create a PR.

ovflowd:
> It looks like the latest jobs of node-test-commit-linuxone ran successfully?

We made a workaround to make it work 🙏

PR #1217, opened 2026-04-14
Add prebuilt binaries for ppc64le and s390x

eelgaev:
GitHub Actions doesn't have ppc64le or s390x runners, so this cross-compiles from x86_64 using the GNU cross toolchains. The linker env vars are set at the job level so both the node bindings build and the CLI build pick them up.

On the npm side, the two new triples are added to build-npm.js (with a powerpc64le → ppc64 arch mapping for Node.js). node/index.js already handles these dynamically, so no changes needed there.

[ppc64 (ppc64le) github build](https://github.com/eelgaev/lightningcss/actions/runs/24375300000/job/71187466385)
[s390x github build](https://github.com/eelgaev/lightningcss/actions/runs/24375300000/job/71187466347)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -174,6 +174,61 @@ jobs:
             *.node
             lightningcss
 
+  build-cross-ppc64le-s390x:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: powerpc64le-unknown-linux-gnu
+            strip: powerpc64le-linux-gnu-strip
+            setup: |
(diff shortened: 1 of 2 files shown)


eelgaev:
Existing issue #1163

eelgaev:
Hi folks, @devongovett, @yisibl, @kdy1, @mischnic, 
As there is more and more adoption for lightningcss, I was hoping you can review this PR, it would really be cool to be able to execute this on ppc64le machines :) 

Thanks!

PR #1218, opened 2026-04-15
feat: Added support for container-type: anchored and @container anchored()

### LINKED ISSUE #1176: Support container-type: anchored and @container anchored()
## Summary

`container-type: anchored` (Chrome 143+) and `@container anchored(fallback: ...)` fail to parse.

## Input

```css
.popover {
  container-type: anchored;
  position-try-fallbacks: flip-block;
}

@container anchored(fallback: flip-block) {
  .popover-arrow {
    inset-block-start: auto;
    inset-block-end: calc(-0.5rem + 1px);
    clip-path: polygon(50% 100%, 0 0, 100% 0);
  }
}
```

## Error

```
Unexpected token Function("anchored")
```

## Context

`container-type: anchored` enables anchor fallback detection on positioned elements. `@container anchored(fallback: <try-fallback>)` queries which `position-try-fallbacks` option was chosen, allowing styles (e.g. arrow direction) to adapt.

- Spec: https://drafts.csswg.org/css-anchor-position-1/#container-queries
- Chrome status: https://chromestatus.com/feature/5196024964833280
- Shipping in Chrome 143

This blocks all Turbopack-based Next.js apps from using anchored container queries since Turbopack uses lightningcss for CSS parsing.

lubomirblazekcz:
fixes #1176

### DIFF
--- a/src/properties/contain.rs
+++ b/src/properties/contain.rs
@@ -2,6 +2,7 @@
 
 #![allow(non_upper_case_globals)]
 
+use bitflags::bitflags;
 use cssparser::*;
 use smallvec::SmallVec;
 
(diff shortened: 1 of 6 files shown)


src/properties/contain.rs
```diff
@@ -11,27 +11,144 @@ use crate::{
...
-    /// Establishes a query container for container scroll-state queries
-    ScrollState,
+/// A value for the [container-type](https://drafts.csswg.org/css-contain-3/#container-type) property.
+/// Establishes the element as a query container for the purpose of container queries.
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "visitor", derive(Visit))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "into_owned", derive(static_self::IntoOwned))]
+pub enum ContainerType {
```
devongovett: A better way to define this would be with bitflags, so that multiple container types can be combined without listing out all of the combinations

src/rules/container.rs
```diff
@@ -195,6 +198,91 @@ impl FeatureToCss for ScrollStateFeatureId {
...
+  derive(serde::Serialize, serde::Deserialize),
+  serde(rename_all = "camelCase")
+)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+pub struct AnchoredFeature<'i> {
+  /// The feature name.
+  pub name: AnchoredFeatureName,
+  /// The feature value.
+  #[cfg_attr(feature = "serde", serde(borrow, with = "ValueWrapper::<TokenList>"))]
+  pub value: TokenList<'i>,
```
devongovett: Ideally we would parse the feature value instead of storing tokens. It should use the same parsing logic/types as for the `position-try-fallbacks` property.

lubomirblazekcz:
@devongovett thanks for the review, I've made the changes with Claude

Anoesj:
It would be great if this could be merged. As tools like Vite now default to lightningcss, this blocks adoption of modern CSS techniques.

lubomirblazekcz:
@devongovett I did one last check/refactor and updated from master, is this good for merge now?

PR #1233, opened 2026-04-30
Print unknown at-rules in the same location

### LINKED ISSUE #1232: Unknown at-rules, are incorrectly hoisted
Right now Lightning CSS doesn't have support for the native `@apply` at-rule (https://drafts.csswg.org/css-mixins-1/#apply-rule). Adding proper support for everything related to this might be a big job. Today, Lightning CSS marks this as an unknown at-rule, but it also hoists the rule to the top which I think is incorrect.

I would expect that unknown at-rules stay in place. However, I'm just thinking in scope of this `@apply` at-rule, so there could be situations where you _do_ want to hoist.

Let's say you have this input:
```css
.foo {
  text-decoration-line: underline;
  @apply --my-mixin-1;
  @apply --my-mixin-1();
  @apply --my-mixin-1 --my-mixin-2;
  @apply --my-mixin-1() --my-mixin-2();
  @apply --my-mixin-3 {
    color: red;
  }
}
```

Lightning CSS marks all of them as unknown, which is fine, but then also hoists them outside of the `.foo` rule:
```css
.foo {
  text-decoration-line: underline;
}

@apply --my-mixin-1;

@apply --my-mixin-1();

@apply --my-mixin-1 --my-mixin-2;

@apply --my-mixin-1() --my-mixin-2();

@apply --my-mixin-3 {
  color: red;
}
```

https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.foo%20%7B%5Cn%20%20text-decoration-line%3A%20underline%3B%5Cn%20%20%40apply%20--my-mixin-1%3B%5Cn%20%20%40apply%20--my-mixin-1()%3B%5Cn%20%20%40apply%20--my-mixin-1%20--my-mixin-2%3B%5Cn%20%20%40apply%20--my-mixin-1()%20--my-mixin-2()%3B%5Cn%20%20%40apply%20--my-mixin-3%20%7B%5Cn%20%20%20%20color%3A%20red%3B%5Cn%20%20%7D%5Cn%7D%5Cn%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D

I think that in this situation, the `@apply` should just stay in-place instead of getting hoisted.

I'll open a PR where it keeps this in-place, but I'll let you decide whether you think this is the correct behavior or not.

RobinMalfait:
This PR is an attempt in fixing #1232 where unknown at-rules were hoisted outside of its parent instead of staying in place. This PR is not adding proper support for `@apply`, but is just used as an example.

I added some tests with various syntaxes as seen in https://drafts.csswg.org/css-mixins-1/#apply-rule. I also added some nesting to make sure that works as expected.

Added some tests with the following input:
```css
.foo {
  @apply --mixin;
  @apply --mixin-no-args();
  @apply --mixin-args(1, 2);
  @apply --mixin-with-empty-body {};
  @apply --mixin-with-body {
    color: red;
  }
  .bar {
    @apply --nested;
  }
}
```

Before, this would generate:
```css
@apply --mixin;

@apply --mixin-no-args();

@apply --mixin-args(1, 2);

@apply --mixin-with-empty-body {
  
}

@apply --mixin-with-body {
  color: red;
}

@apply --nested;
```
As you can see, the `.foo` and `.bar` are even gone, because at some point they won't contain any declarations anymore and are just removed.

But with this PR, it will generate:
```css
.foo {
  @apply --mixin;
  @apply --mixin-no-args();
  @apply --mixin-args(1, 2);
  @apply --mixin-with-empty-body {
    
  }
  @apply --mixin-with-body {
    color: red;
  }
}

.foo .bar {
  @apply --nested;
}
```

Like I mentioned before, this doesn't add proper support for `@apply`, so this is just an attempt in properly printing the correct output. There is no validation related to the actual mixins that are being applied here.

Fixes: https://github.com/parcel-bundler/lightningcss/issues/1232

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24230,6 +24230,58 @@ mod tests {
     nesting_test(
       r#"
         .foo {
+          @apply --mixin;
+          @apply --mixin-no-args();
+          @apply --mixin-args(1, 2);
+          @apply --mixin-with-empty-body {};
+          @apply --mixin-with-body {
+            color: red;
+          }
+          .bar {
+            @apply --nested;
+          }
+        }
+      "#,
+      indoc! {r#"
+        .foo {
+          @apply --mixin;
+          @apply --mixin-no-args();
+          @apply --mixin-args(1, 2);
+          @apply --mixin-with-empty-body {
+            
+          }
+          @apply --mixin-with-body {
+            color: red;
+          }
+        }
+
+        .foo .bar {
+          @apply --nested;
+        }
+      "#},
+    );
+
+    minify_test(
+      r#"
+        .foo {
+          @apply --mixin;
+          @apply --mixin-no-args();
+          @apply --mixin-args(1,2);
+          @apply --mixin-with-empty-body {};
+          @apply --mixin-with-body {
+            color: red;
+          };
+          .bar {
+            @apply --nested;
+          }
+        }
+      "#,
+      ".foo{@apply --mixin;@apply --mixin-no-args();@apply --mixin-args(1,2);@apply --mixin-with-empty-body{}@apply --mixin-with-body{color: red;}& .bar{@apply --nested;}}",
+    );
+
+    nesting_test(
+      r#"
+        .foo {
           color: blue;
           & > .bar { color: red; }
         }
--- a/src/rules/style.rs
+++ b/src/rules/style.rs
@@ -11,7 +11,7 @@ use crate::error::ParserError;
 use crate::error::{MinifyError, PrinterError};
 use crate::parser::DefaultAtRule;
 use crate::printer::Printer;
-use crate::rules::CssRuleList;
+use crate::rules::{CssRule, CssRuleList};
 use crate::selector::{
   downlevel_selectors, get_prefix, is_compatible, is_pure_css_modules_selector, is_unused, SelectorList,
 };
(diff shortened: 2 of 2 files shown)


devongovett (approved):
I guess this makes sense. With other at rules, the at rule gets hoisted and the style rule gets placed inside it, but I guess if the at rule is unknown we can't do that or even know if it's the desired behavior so just preserving it as is makes sense.

Issue #660, opened 2024-01-16
feat: CSS Modules `[hash]` to include class name


privatenumber:
## Feature request

I'd like to request for the CSS Modules `[hash]` to include the class name. This would allow `cssModules.pattern` to be `[hash]` alone, without needing `[local]`.

## Why?

After CSS modularization, I use CSS Modules for these two major benefits:

1. It condenses long, verbose class names into shorter, more manageable ones.

2. It obfuscates class names, helping prevent reverse-engineering tactics, such as crawlers using selectors.

Furthermore, coming from a background of using CSS Modules with Webpack's `css-loader`, where [`[hash]` generates a hash unique to each class name](https://webpack.js.org/loaders/css-loader/#modules:~:text=%5Bhash%5D%20%2D%20the%20hash%20of%20the%20string%2C%20generated%20based%20on%20localIdentHashSalt%2C%20localIdentHashFunction%2C%20localIdentHashDigest%2C%20localIdentHashDigestLength%2C%20localIdentContext%2C%20resourcePath%20and%20exportName), I found a slight migration incompatibility with the current implementation in LightningCSS.

P.S. Great work on LightningCSS! It's packed with impressive features that I really appreciate.

sciyoshi:
#1237 should cover this. There's a new `Config::hash_local_name` opt-in (off by default) that appends the local class name to the hash input, so `[hash]` becomes unique per class — matching css-loader's behavior. Works with the new `[<algo>:hash:<digest>:<length>]` placeholders too if you want md4/base64/short-length output.

Issue #636, opened 2023-11-24
Match CSS modules hashing with PostCSS


marvinhagemeister:
I'm currently exploring adding support for CSS Modules natively in Deno via lightningcss. In doing so I noticed that the hash segment is a different value then the one calculated by PostCSS. It seems like PostCSS is using a hex digest of the `xxhash64` hash of the root relative file path compared to lightningcss which uses rust's default `SipHash-1-3` algorithm.

To make native CSS Modules support in Deno useful, the hash value needs to match whatever users may be using to generate browser bundles with. Would folks be open to the idea to change the hashing function here to match the one from PostCSS? I'd be happy to make a PR for that.

devongovett:
Why would the exact hash matter? Isn't the point of hashing css modules so that they contain an unpredictable value that won't conflict?

SevereCloud:
For example, to use with github.com/VKCOM/swc-plugin-css-modules

sciyoshi:
#1237 takes a swing at this. It lands the four pieces needed for byte-parity with PostCSS / css-loader as independent opt-in flags on `Config` — webpack-style algo/digest/length placeholders, hash salt, per-local hashing, and a post-process pass for ident validity. Verified against captured ground-truth output from a real Vite build. Defaults preserve existing lightningcss output.

Issue #491, opened 2023-05-18
More flexibility over cssModules pattern


mayank99:
Currently, the generated scoped class can be customized to an extent:
```js
cssModules: {
  pattern: 'whatever-[name]-[hash]-[local]'
}
```

It would be nice if there was a way to customize it further. One of the use cases for this is when file names / local classes are too long and need to be truncated.

This could be offered using a function (might be difficult to communicate into Rust):
```js
cssModules: {
  pattern: (name, hash, local) => `w-${name.split('-')[0]}-${hash}-${local.slice(0, 5)}'
}
```

Or it could be through more advanced pattern matching that includes built-in truncation (kinda similar to [webpack](https://github.com/webpack/loader-utils#interpolatename)).
```js
cssModules: {
  pattern: 'w-[name:5]-[hash]-[local:4]'
}
```

sciyoshi:
#1237 covers the `[hash]` truncation half of this — patterns like `[hash:5]`, `[md4:hash:base64:5]`, etc. now parse. Doesn't help with `[name:5]` / `[local:4]` though; those would be a separate change.

Issue #351, opened 2022-12-07
CSS Modules hashes clash within the same file


markdalgleish:
I ran into an issue where styles worked in dev but broke in the production build because I set the `pattern` option to `[hash]` but it turned out that the hashes were the same for each `.module.css` file.

You can see this in the current test suite. Notice how the classes are only unique because of the trailing `_[local]` value. This would break if only the hashes were included: https://github.com/parcel-bundler/lightningcss/blob/27cb9c2a382fd8d9702de8def9dd18db54280c43/tests/cli_integration_tests.rs#L95-L129

devongovett:
At the moment `[hash]` is just a hash of the file path. I guess we could add a new pattern option to hash the whole class, or maybe change the behavior, but I avoided this initially for performance reasons.

markdalgleish:
Could the hash include a counter at the end? This is what we do in Vanilla Extract, so in this example we'd end up with `EgL3uq0`, `EgL3uq1` etc.

markdalgleish:
Ah, that's interesting. It's also a bit of a gotcha that the build suddenly breaks if I start using grid line names when I have a pattern that doesn't end in `[local]`.

devongovett:
Yeah, might add some validation of the config to ensure that it will generate unique, valid names.

The counter at the end is an interesting idea, but might not be stable over time e.g. if you insert a new rule. Will continue to think through options.

sciyoshi:
Ran into the same surprise. #1237 should fix this with an opt-in `Config::hash_local_name` flag — when set, the local class name is mixed into the hash input so classes within the same file get distinct hashes. Off by default to preserve existing output, but worth a discussion about whether to flip the default in a future major to prevent this footgun.

PR #1237, opened 2026-05-05
CSS modules: webpack/css-loader-compatible scoped-name hashing

### LINKED ISSUE #503: Is it possible to change hash function for css modules?
Hi, i try to use lightningcss in javascript module loader.

It is working but i need to specify hash function.

How can i do it?

sciyoshi:
Adds opt-in support for producing CSS-modules scoped names byte-identical to webpack/css-loader and Vite-via-postcss-modules output. Default `[hash]` behavior is unchanged.

Reverse-engineered the algorithm from `loader-utils.getHashDigest` + `loader-utils.interpolateName` + `genericNames` against ground-truth output captured from a real Vite/postcss-modules build. End-to-end byte-parity verified across 12 captured class names exercising every code path (including the leading-digit / leading-`-digit` / leading-`--` prefix branches).

Closes #503, #636, #660, #351. Partially addresses #491 (`[hash]` truncation only).

## New pattern syntax

Webpack-style `[<algo>:hash:<digest>:<length>]` placeholders. Each part is independently optional; the position of the literal `hash` keyword distinguishes algo (before) from digest+length (after):

| Pattern | Algorithm | Digest | Length |
|---|---|---|---|
| `[hash]` | legacy lightningcss | legacy | n/a |
| `[hash:base64]` | xxhash64 (default) | base64 | full |
| `[hash:5]` | xxhash64 | hex (default) | 5 |
| `[md4:hash:base64:5]` | md4 | base64 | 5 |
| `[xxhash64:hash:hex:12]` | xxhash64 | hex | 12 |

Bare `[hash]` parses to all-`None` and stays on the legacy code path — existing snapshots are byte-identical. Parsing is case-insensitive. Unknown algo, unknown digest, missing `hash` keyword, and extra parts before `hash` all return `PatternParseError::UnknownPlaceholder`.

`Segment::ContentHash` is intentionally left as a unit variant in this PR — same options on `[content-hash]` would require plumbing through the bundler's content-hash precomputation and is a follow-up.

## New `Config` fields

Three new opt-in fields on `css_modules::Config`, all defaulting to no-op so existing output is unchanged:

| Field | Default | Effect when set |
|---|---|---|
| `hash_prefix: Option<Cow<'i, str>>` | `None` | Prepended to every hash input. `"\x00\x00\x00\x00"` reproduces css-loader's tier-0 salt input byte-for-byte. |
| `hash_local_name: bool` | `false` | Appends the local name (separated by NUL) to each hash input. Matches css-loader/postcss-modules per-(file, local) hashing rather than lightningcss's per-file hashing. |
| `escape_scoped_names: bool` | `false` | Post-processes every rendered scoped name through `loader-utils`' `genericNames` rules: replace any char outside `[a-zA-Z0-9\-_ -￿]` with `-`, then prefix `_` when the result starts with `-?[0-9]` or `--`. Required for standard-base64 digests like `[md4:hash:base64:5]` to produce valid CSS identifiers. |

Each is independently meaningful, but for css-loader byte-parity they travel together with a webpack-style pattern. The byte-parity end-to-end test in `tests::css_modules_webpack_parity` exercises all four together against the captured Vite output.

## Public API additions

- `pub enum HashAlgorithm { Md4, Xxhash64 }`
- `pub enum DigestType { Hex, Base64 }`
- `Segment::Hash` becomes a struct variant: `{ algo: Option<HashAlgorithm>, digest: Option<DigestType>, length: Option<usize> }`
- `Config { hash_prefix, hash_local_name, escape_scoped_names }`
- `pub(crate) fn hash_with_options(input: &[u8], algo, digest, length) -> String` (used by both the new pattern path and the byte-parity test)

## JS API additions

`CSSModulesConfig` in `node/index.d.ts` and `napi::CssModulesConfig` gain three optional camelCase fields (`hashPrefix`, `hashLocalName`, `escapeScopedNames`) that thread through to the Rust `Config`. Defaults preserve existing JS-side behavior.

This is what makes the feature reachable from Vite via `css.lightningcss.cssModules`.

## Why

CSS-modules ecosystem hashing — used by webpack/css-loader, postcss-modules, and Vite-with-PostCSS — is genuinely opinionated, but it's also the de facto convention. Lightningcss's existing `[hash]` is more lightweight (per-file, custom alphabet, no salt, no post-processing) which is fine for a fresh project, but doesn't let anyone migrate from PostCSS to lightningcss without changing every scoped name in the build. That's a hard sell for production codebases.

The four pieces here are deliberately independent flags rather than a single `compat_mode` bundle, on the principle that the API should be honest about which knob does what. A user who only needs salting can enable `hash_prefix` without buying into per-local hashing or escaping. For full webpack parity all four go on at once.

## Backward compatibility

- Default `[hash]` is unchanged. The legacy code path is preserved when `algo`, `digest`, and `length` are all `None`.
- All three new `Config` fields default to `None` / `false`. No existing user sees different output until they opt in.
- Existing tests pass byte-identical (verified locally — `cargo test --no-default-features` shows only the pre-existing platform-dependent `tests::test_dependencies` SipHash snapshot failure, unrelated to this work).

## Test plan

- [x] `cargo test --no-default-features` — 135 pass
- [x] **Byte-parity end-to-end test** (`tests::css_modules_webpack_parity`) reproduces 12 captured Vite/postcss-modules class names exactly. Cases cover every post-process branch:
  - Clean digest (`Alpha-module__foo__YTbdH`)
  - `+` and `/` cleanup (`Alpha-module__cls_4__LOY-5`, `Alpha-module__cls_48__-ta-0`)
  - Leading-digit prefix (digest leads → `_2kP0r__…`)
  - Leading `-digit` prefix (digest `+0F5/` → `-0F5-` → `_-0F5-__…`)
  - Leading `--` prefix (digest `//GdO` → `--GdO` → `_--GdO__…`)
  - Leading `-letter` (no prefix)
- [x] 26 new unit tests in `css_modules::tests` covering parser forms, rejection cases, hash dispatch, and each post-process branch in isolation

## Commits

1. `chore(css-modules): add md4 and xxhash-rust dependencies`
2. `feat(css-modules): add HashAlgorithm and DigestType with hash_with_options helper`
3. `feat(css-modules): extend Segment::Hash with optional algo, digest, and length`
4. `feat(css-modules): parse webpack-style [<algo>:hash:<digest>:<length>] syntax`
5. `feat(css-modules): add Config::hash_prefix for hash input salting`
6. `feat(css-modules): add Config::hash_local_name for per-local hashing`
7. `feat(css-modules): add Config::escape_scoped_names post-process pipeline`
8. `test(css-modules): add Vite scoped-name byte-parity end-to-end test`
9. `feat(napi): expose hash_prefix, hash_local_name, escape_scoped_names to JS`

### DIFF
--- a/node/index.d.ts
+++ b/node/index.d.ts
@@ -313,7 +313,11 @@ export interface Warning {
 }
 
 export interface CSSModulesConfig {
-  /** The pattern to use when renaming class names and other identifiers. Default is `[hash]_[local]`. */
+  /**
+   * The pattern to use when renaming class names and other identifiers. Default is `[hash]_[local]`.
+   * Extended hash patterns such as `[md4:hash:base64:5]` automatically apply legacy
+   * css-loader/postcss-modules scoped-name post-processing.
+   */
   pattern?: string,
   /** Whether to rename dashed identifiers, e.g. custom properties. */
   dashedIdents?: boolean,
@@ -340,14 +344,6 @@ export interface CSSModulesConfig {
    * Default: false (lightningcss hashes once per file).
    */
   hashLocalName?: boolean,
-  /**
-   * Run css-loader/postcss-modules' `genericNames` post-process on every rendered
-   * scoped name: replace any char outside `[a-zA-Z0-9\-_ -￿]` with `-`,
-   * and prefix `_` when the result starts with `-?[0-9]` or `--`. Required for
-   * standard-base64 digests like `[md4:hash:base64:5]` to produce valid CSS
-   * identifiers. Default: false.
-   */
-  escapeScopedNames?: boolean
 }
 
 export type CSSModuleExports = {
--- a/node/index.d.ts
+++ b/node/index.d.ts
@@ -336,7 +336,7 @@ export interface CSSModulesConfig {
   hashPrefix?: string,
   /**
    * Append the local class/ident name (separated by NUL) to each hash input.
-   * Required to match css-loader/postcss-modules' per-(file, local) hashing.
+   * Required to reproduce legacy css-loader/postcss-modules scoped names.
    * Default: false (lightningcss hashes once per file).
    */
   hashLocalName?: boolean,
--- a/node/index.d.ts
+++ b/node/index.d.ts
@@ -339,7 +339,8 @@ export interface CSSModulesConfig {
    */
   hashPrefix?: string,
   /**
-   * Append the local class/ident name (separated by NUL) to each hash input.
+   * Append the local class/ident name (separated by NUL) to each `[hash]` input.
+   * This is separate from `[content-hash]`, which is derived from file contents.
    * Required to reproduce legacy css-loader/postcss-modules scoped names.
    * Default: false (lightningcss hashes once per file).
    */
--- a/node/index.d.ts
+++ b/node/index.d.ts
@@ -326,7 +326,28 @@ export interface CSSModulesConfig {
   /** Whether to enable hashing for custom identifiers. */
   customIdents?: boolean,
   /** Whether to require at least one class or id selector in each rule. */
-  pure?: boolean
+  pure?: boolean,
+  /**
+   * Prepend this string to the bytes hashed for every `[hash]` segment. Mirrors
+   * Vite/postcss-modules' `hashPrefix` option and css-loader's `hashSalt` (when
+   * set to "\x00\x00\x00\x00", lightningcss reproduces css-loader's tier-0 salt
+   * input byte-for-byte). Default: empty.
+   */
+  hashPrefix?: string,
+  /**
+   * Append the local class/ident name (separated by NUL) to each hash input.
+   * Required to match css-loader/postcss-modules' per-(file, local) hashing.
+   * Default: false (lightningcss hashes once per file).
+   */
+  hashLocalName?: boolean,
+  /**
+   * Run css-loader/postcss-modules' `genericNames` post-process on every rendered
+   * scoped name: replace any char outside `[a-zA-Z0-9\-_ -￿]` with `-`,
+   * and prefix `_` when the result starts with `-?[0-9]` or `--`. Required for
+   * standard-base64 digests like `[md4:hash:base64:5]` to produce valid CSS
+   * identifiers. Default: false.
+   */
+  escapeScopedNames?: boolean
 }
 
 export type CSSModuleExports = {
--- a/src/css_modules.rs
+++ b/src/css_modules.rs
@@ -58,10 +58,12 @@ pub struct Config {
   /// no effect on `[content-hash]`. Default is `None` (no prefix; Lightning CSS output is
   /// unchanged from prior versions).
   pub hash_prefix: Option<Cow<'static, str>>,
-  /// When `true`, the local class/ident name is appended to the hash input separated by
-  /// a NUL byte: `<prefix><relative-path>\0<local>`. This matches the per-local hashing
-  /// done by css-loader and postcss-modules — without it, every export from a given file
-  /// shares one hash. Required to reproduce legacy css-loader/postcss-modules scoped names.
+  /// When `true`, the local class/ident name is appended to the `[hash]` input separated
+  /// by a NUL byte: `<prefix><relative-path>\0<local>`. This matches the per-local
+  /// hashing done by css-loader and postcss-modules — without it, every export from a
+  /// given file shares one `[hash]`. This is separate from `[content-hash]`, which is
+  /// derived from file contents rather than from each local name. Required to reproduce
+  /// legacy css-loader/postcss-modules scoped names.
   ///
   /// Default is `false` (preserves lightningcss's per-file hashing behavior).
   pub hash_local_name: bool,
--- a/src/css_modules.rs
+++ b/src/css_modules.rs
@@ -58,6 +58,13 @@ pub struct Config {
   /// no effect on `[content-hash]`. Default is `None` (no prefix; lightningcss output is
   /// unchanged from prior versions).
   pub hash_prefix: Option<Cow<'static, str>>,
+  /// When `true`, the local class/ident name is appended to the hash input separated by
+  /// a NUL byte: `<prefix><relative-path>\0<local>`. This matches the per-local hashing
+  /// done by css-loader and postcss-modules — without it, every export from a given file
+  /// shares one hash. Required for css-loader/postcss-modules byte parity.
+  ///
+  /// Default is `false` (preserves lightningcss's per-file hashing behavior).
+  pub hash_local_name: bool,
 }
 
 impl Default for Config {
@@ -71,6 +78,7 @@ impl Default for Config {
       custom_idents: true,
       pure: false,
       hash_prefix: None,
+      hash_local_name: false,
     }
   }
 }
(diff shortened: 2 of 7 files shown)


devongovett:
> but doesn't let anyone migrate from PostCSS to lightningcss without changing every scoped name in the build

Not necessarily opposed to adding these, but isn't the whole point of CSS modules that the scoped names are not predictable?

node/index.d.ts
```diff
@@ -321,7 +321,28 @@ export interface CSSModulesConfig {
...
+   */
+  hashLocalName?: boolean,
+  /**
+   * Run css-loader/postcss-modules' `genericNames` post-process on every rendered
+   * scoped name: replace any char outside `[a-zA-Z0-9\-_ -￿]` with `-`,
+   * and prefix `_` when the result starts with `-?[0-9]` or `--`. Required for
+   * standard-base64 digests like `[md4:hash:base64:5]` to produce valid CSS
+   * identifiers. Default: false.
+   */
+  escapeScopedNames?: boolean
```
devongovett: Why is this an option? Won't `false` produce invalid CSS? Shouldn't it default to true?

src/css_modules.rs
```diff
@@ -296,92 +432,128 @@ impl<'a, 'b, 'c> CssModule<'a, 'b, 'c> {
...
+        &self.sources[source_index as usize],
+        local,
+        if let Some(content_hashes) = &self.content_hashes {
+          &content_hashes[source_index as usize]
+        } else {
+          ""
+        },
+      )
+      .unwrap();
+    let name = self.maybe_escape(body);
```
devongovett: Can we keep this logic inside the `or_insert_with` so we only do it if the entry doesn't already exist?

src/css_modules.rs
```diff
@@ -296,92 +432,128 @@ impl<'a, 'b, 'c> CssModule<'a, 'b, 'c> {
...
+        &self.sources[source_index as usize],
+        &local[2..],
+        if let Some(content_hashes) = &self.content_hashes {
+          &content_hashes[source_index as usize]
+        } else {
+          ""
+        },
+      )
+      .unwrap();
+    let name = format!("--{}", self.maybe_escape(body));
```
devongovett: same

sciyoshi:
> Not necessarily opposed to adding these, but isn't the whole point of CSS modules that the scoped names are not predictable?

Indeed. However, in the real world there are situations where things outside of one's control depend on these classnames. In our case specifically, we have an extensive automated test suite which relies on these classnames to select elements. While we're doing work in parallel to remove that reliance on these mangled names, this approach is more straightforward and unblocks us from switching to LightningCSS.

src/css_modules.rs
```diff
@@ -546,3 +725,342 @@ pub(crate) fn hash(s: &str, at_start: bool) -> String {
...
+  Md4,
+  /// xxHash64. Matches webpack's default `xxhash64` hash function.
+  Xxhash64,
+}
+
+/// Post-process a rendered scoped name using css-loader/postcss-modules' `genericNames`
+/// rules: replace any char outside `[a-zA-Z0-9\-_]` (plus the latin-1+ unicode range
+/// `U+00A0..=U+FFFF`) with `-`, then prefix `_` when the result starts with `-?[0-9]`
+/// or `--` so the output remains a valid CSS identifier.
+pub(crate) fn escape_scoped_name(s: &str) -> String {
```
devongovett, line 718: The Printer already uses `serialize_identifier` / `serialize_name` in its `write_ident` method, which is what calls `css_module.config.pattern.write`. Those functions already escape the name, so I don't think we need to implement a custom escaping function here as well?

node/index.d.ts
```diff
@@ -321,7 +321,28 @@ export interface CSSModulesConfig {
...
+   * set to "\x00\x00\x00\x00", lightningcss reproduces css-loader's tier-0 salt
+   * input byte-for-byte). Default: empty.
+   */
+  hashPrefix?: string,
+  /**
+   * Append the local class/ident name (separated by NUL) to each hash input.
+   * Required to match css-loader/postcss-modules' per-(file, local) hashing.
+   * Default: false (lightningcss hashes once per file).
+   */
+  hashLocalName?: boolean,
```
devongovett, line 347: Lightningcss already supports `[contenthash]` which does this. Seems confusing to have two ways of doing the same thing?

src/css_modules.rs
sciyoshi: Good catch, fixed in `d112c8d`.

Looking at this again, the printer had already rendered the scoped name it writes to CSS, and then `add_local` rendered the same name again for the exports map. I changed `add_local` to take the printer’s already-computed scoped name and use `or_insert_with`, so the helper no longer repeats the pattern rendering work.

src/css_modules.rs
sciyoshi: Fixed this one in the same commit, `d112c8d`.

`write_dashed_ident` now passes the already-computed scoped custom-property name into `add_dashed`, and `add_dashed` uses `or_insert_with`. While I was there, I also cleaned up the local `reference_dashed` path so it checks for an existing entry before rendering a new scoped name.

src/css_modules.rs
sciyoshi: These are doing different things, but I should have called that out. Added a doc comment in `ea0243f`.

`serialize_identifier` / `serialize_name` make a string valid CSS syntax by backslash-escaping at print time. `escape_scoped_name` rewrites the literal scoped name to match legacy css-loader/postcss-modules output.

For example, serializer escaping would make `foo+bar` printable as `foo\+bar` in CSS, but the JS exports map would still have the literal `foo+bar`. The compatibility post-process instead rewrites both sides to the same literal name, e.g. `foo-bar`. The printer still serializes the rewritten name afterward for general CSS safety.

node/index.d.ts
sciyoshi: I see why these look related, but they’re solving different problems. I clarified the docs in `4f76fcc`.

`[content-hash]` is derived from file contents, so every export in a given module sees the same content hash. `hashLocalName` changes the input to the `[hash]` segment to include the local export name: `<prefix><path>\0<local>`. That means each export in the same file can get a different hash, which is what css-loader/postcss-modules do.

sciyoshi:
Thanks so much for taking the time to review this; much appreciated, especially given the size of the diff.

I pushed a few small follow-up commits addressing the inline comments:
- `0791502` clarifies the “legacy” terminology so Lightning CSS’s existing behavior is described as the default/native path, and legacy compatibility refers to webpack/css-loader/postcss-modules.
- `baffdb0` removes `escapeScopedNames` as a public option and applies the compatibility post-process automatically for extended hash patterns.
- `d112c8d` avoids re-rendering scoped names in `add_local` / `add_dashed`.
- `ea0243f` documents why `escape_scoped_name` is distinct from CSS serializer escaping.
- `4f76fcc` clarifies how `hashLocalName` differs from `[content-hash]`.

I also rebased this on the current master branch to fix the merge conflict around CSS module pattern ownership (master moved `Pattern` / `Segment` to an owned model without lifetimes).

Separately: full webpack/css-loader parity was the bar we wanted for our own migration. I completely understand if that scope is more than you want to maintain long-term. If any of this feels like too much new config or compatibility behavior to maintain, we’re happy to keep it in a fork for our internal use and you can take only the pieces that feel useful to LightningCSS, or none of it. No pressure to merge as-is; your call, and thanks again for the thoughtful review.

node/index.d.ts
sciyoshi: Agreed. I removed `escapeScopedNames` as a public option instead of just flipping the default.

The post-process now runs automatically whenever the pattern uses the extended legacy webpack/css-loader-compatible hash syntax, e.g. `[md4:hash:base64:5]`. Bare `[hash]` still keeps Lightning CSS’s default hash path unchanged.

Added a regression test asserting extended/base64 hash patterns are escaped automatically. Done in `baffdb0`.

PR #1240, opened 2026-05-13
feat: refine declarations error recovery

intellild:
## Summary

  - Refine CSS error recovery inspired by csstree's fallback-to-raw parsing model.
  - Add a `Raw` AST node for recovered invalid CSS fragments.
  - Preserve invalid rule/declaration text as `Raw` only when `error_recovery` is enabled.
  - Keep existing strict parsing behavior when `error_recovery` is disabled.
  - Recover valid selectors from invalid comma-separated selector lists, e.g. `invalid@selector, .green` now preserves `.green`.
  - Add regression coverage for invalid selectors and invalid declarations.
  - Ignore local `.idea` IDE settings.

  ## Validation

  - `cargo test`
  - `cargo test test_error_recovery --lib`


devongovett:
What's the real-world use case for this?

intellild:
> What's the real-world use case for this?

Currently, Lightning CSS's error recovery capabilities are far from the CSS standards and browser implementations, resulting in significant differences in the output of many CSS files. Consequently, the styles that actually take effect in the browser may differ from expectations, especially for third-party CSS libraries. 
This experimental PR aims to enhance the parser's error recovery capabilities to improve the developer experience. 
My future plan is to implement more PostCSS plugin features in Lightning CSS, so that it can fully replace PostCSS in real-world projects.

devongovett:
As I wrote [here](https://github.com/parcel-bundler/lightningcss/discussions/1241) under project philosophy, it is not a goal of the project to support non-standard syntax extensions like PostCSS does. Lightning CSS is focused on supporting CSS standards. PostCSS is still a great tool for supporting custom syntaxes. Currently, Lightning CSS implements error recovery exactly how browsers do - by ignoring invalid rules.

Could you describe in more detail what kind of syntax errors you are looking to pass through and how the current behavior differs from browsers?

intellild:
> As I wrote [here](https://github.com/parcel-bundler/lightningcss/discussions/1241) under project philosophy, it is not a goal of the project to support non-standard syntax extensions like PostCSS does. Lightning CSS is focused on supporting CSS standards. PostCSS is still a great tool for supporting custom syntaxes. Currently, Lightning CSS implements error recovery exactly how browsers do - by ignoring invalid rules.
> 
> Could you describe in more detail what kind of syntax errors you are looking to pass through and how the current behavior differs from browsers?

I currently don’t have plans to support non-standard syntax. This change was only an experiment to improve the user experience by preserving invalid code fragments in the output and evaluating the cost of doing so. As you can see, all related changes are limited to development mode.

I’ll close this for now. Thanks for your understanding.

PR #1250, opened 2026-05-23
Don’t deduplicate `font-family: monospace, monospace`

### LINKED ISSUE #1221: `font-family: monospace, monospace` should not deduplicate (the duplication is load-bearing)
[Playground](https://lightningcss.dev/playground/index.html#{"source":"code{font-family:monospace,monospace}"})

Input:
```css
code{fony-family:monospace,monospace}
```

Output (incorrect, changes rendering in most browsers):
```css
code{fony-family:monospace}
```

----

Not sure what’s going on? It’s a stupid piece of historical baggage in browsers: `monospace`, *when it’s the only font in the family stack*, defaults to probably 13px, when all other text is probably 16px. Can’t find this behaviour in any spec, but everyone’s done it since the very beginning. The whole thing was a mistake in my opinion, and people often work around it with `font-size: 1em` (or their own size that probably isn’t 81.25%), but I suspect a lot of developers don’t know about it any more because most people use longer `font-family` declarations (for better or normally for worse).

Well, I like to fix it with `font-family: monospace, monospace`, which is the only correct way of resolving it within `font-family` alone (as any other font name could theoretically alter the rendering, if the `monospace` face is missing a character and the provided name maps to a face other than the default face and which has it).

Unfortunately Lightning CSS removes the duplication. It shouldn’t, *specifically if it’s `monospace` and this would reduce the list to one member*.

I don’t know of this idiosyncrasy of `font-family` affecting anything else, but it conceivably could.

chris-morgan:
~~Unless size is specified too, then it’s safe.~~

~~Monospace is the main place this has visible effects, but it’s conservative to apply it to all Generic; it can theoretically have effects with more, especially if you add languages to the mix.~~

I *believe* `monospace` is the only name where duplication has effects. Different languages can vary sizes as well, but that’s not a CSS concern.

Fixes #1221.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6364,6 +6364,19 @@ mod tests {
       "@font-face{font-family:\"revert-layer\"}",
     );
 
+    minify_test(
+        ".foo{font-family:named-font,named-font}",
+        ".foo{font-family:named-font}",
+    );
+    minify_test(
+        ".foo{font-family:monospace,monospace}",
+        ".foo{font-family:monospace,monospace}",
+    );
+    minify_test(
+        ".foo{font-family:monospace,monospace;font-size:16px}",
+        ".foo{font-family:monospace,monospace;font-size:16px}",
+    );
+
     prefix_test(
       r#"
       .foo {
--- a/src/properties/font.rs
+++ b/src/properties/font.rs
@@ -924,7 +924,14 @@ impl<'i> FontHandler<'i> {
     let variant_caps = std::mem::take(&mut self.variant_caps);
 
     if let Some(family) = &mut family {
-      if family.len() > 1 {
+      if family.iter().all(|f| matches!(f, FontFamily::Generic(GenericFontFamily::Monospace))) {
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -6364,6 +6364,19 @@ mod tests {
       "@font-face{font-family:\"revert-layer\"}",
     );
 
+    minify_test(
+        ".foo{font-family:named-font,named-font}",
+        ".foo{font-family:named-font}",
+    );
+    minify_test(
+        ".foo{font-family:monospace,monospace}",
```
yisibl, line 6372: What if there are three or more?

chris-morgan: I don’t *actually* think that’s a concern—it can only happen on purpose and would be decidedly weird—but I incidentally made it work for more than two as part of other changes.

chris-morgan:
I’ve updated this based on the extra knowledge acquired in the original issue. I think it’s ready.

PR #1258, opened 2026-06-01
feat: add RISC-V (riscv64gc-unknown-linux-gnu) prebuilt binary support

rockyzhang:
Add cross-compilation and npm packaging support for RISC-V:

- .cargo/config.toml: add riscv64gc-unknown-linux-gnu linker
- .github/workflows/release.yml: add RISC-V build matrix entry using ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian image with gcc-riscv64-linux-gnu cross-compiler
- scripts/build-npm.js: register riscv64gc-unknown-linux-gnu triple with glibc and map cpu arch to riscv64

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -7,6 +7,9 @@ linker = "aarch64-linux-gnu-gcc"
 [target.armv7-unknown-linux-gnueabihf]
 linker = "arm-linux-gnueabihf-gcc"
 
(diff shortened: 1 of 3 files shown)


devongovett:
I'm not sure this has wide enough usage to take on officially. We also have no way to test that it works. What hardware are you running this on? Can you just compile it from source?

PR #1259, opened 2026-06-02
fix: preserve `backdrop-filter` when webkit fallback is present

### LINKED ISSUE #695: Backdrop Filter outputs only preffixed version
Hey all!

I'm using lightningcss with pandacss and the output of `backdrop-filter` is only `-webkit-backdrop-filter: '...'`, causing it not to work on the most recent version of Edge.

This is my current browserlist setup: `'last 10 versions', 'not dead', 'not < 2%'`

fireairforce:
closes: 

closes https://github.com/parcel-bundler/lightningcss/issues/695
https://github.com/vercel/next.js/issues/78302
https://github.com/utooland/utoo/issues/3090

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28331,6 +28331,15 @@ mod tests {
       "#,
       ".foo{-webkit-backdrop-filter:blur(10px);backdrop-filter:blur(10px)}",
     );
+    minify_test(
+      r#"
+        .foo {
+          backdrop-filter: blur(10px);
+          -webkit-backdrop-filter: blur(20px);
+        }
+      "#,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -28322,6 +28322,15 @@ mod tests {
         ..Browsers::default()
       },
     );
+    minify_test(
+      r#"
+        .foo {
+          backdrop-filter: blur(10px);
+          -webkit-backdrop-filter: blur(10px);
```
yisibl, line 28329: What happens if the values of prefixed properties are different? It’s best to have a test to cover this scenario.

fireairforce: done

PR #1262, opened 2026-06-03
Allow using functions as patterns

### LINKED ISSUE #491: More flexibility over cssModules pattern

Delta-official:
The Rust `Pattern` is now an enum that can be either a string or an `Arc<dyn PatternProvider>`, the latter of which implements `write` and `needs_content_hash` (which stands in for the code checking for the inclusion of `[content-hash]` in the pattern.

The only real change on the JS side is that you can pass in an object of the form:
```ts
export interface CSSModulePattern {
  /** The function to use to resolve patterns */
  resolve: (hash: string, path: string, local: string, content_hash?: string) => MaybePromise<string>,
  /** Whether to pass in the content_hash param into `resolve`. */
  content_hash: boolean
}
```
instead of a string into the `cssModules.pattern` field

Resolves: #491

### DIFF
--- a/napi/src/lib.rs
+++ b/napi/src/lib.rs
@@ -1,6 +1,7 @@
 #[cfg(feature = "bundler")]
 use at_rule_parser::AtRule;
 use at_rule_parser::{CustomAtRuleConfig, CustomAtRuleParser};
+use crossbeam_channel::{Receiver, Sender};
 use lightningcss::bundler::BundleErrorKind;
 #[cfg(feature = "bundler")]
 use lightningcss::bundler::{Bundler, SourceProvider};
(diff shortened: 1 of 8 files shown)


Delta-official:
@devongovett could you please review this?

PR #1268, opened 2026-06-05
Add riscv64 (linux-gnu) prebuilt binary

### LINKED ISSUE #1267: Add prebuilt riscv64 (linux-gnu) binary
lightningcss ships prebuilt bindings for linux x64/arm64/arm in both gnu and musl, but not riscv64. On a riscv64 box `require('lightningcss')` just throws: the loader looks for `lightningcss-linux-riscv64-gnu` (not on npm), then a bundled `lightningcss.linux-riscv64-gnu.node` (not shipped), and there's no wasm fallback on that path.

riscv64gc-unknown-linux-gnu is a supported Rust target and Node has riscv64 builds now, so the only missing piece is the prebuilt package. Anything that pulls lightningcss in (Parcel, vite, the @tailwindcss/oxide path) currently needs a manual workaround to run on these boards.

I built it both ways to be sure it actually works:

- cross-compiled in the napi-rs container (riscv64gc-unknown-linux-gnu), ~3 min
- natively on a riscv64 runner

Both produce a `lightningcss.linux-riscv64-gnu.node` that loads and minifies fine: `.a{ color: #ffffff }` comes out as `.a{color:#fff}`.

If you're open to it, I'm happy to send a PR adding `riscv64gc-unknown-linux-gnu` to the release matrix. It follows the existing cross entries: same napi-rs image plus `gcc-riscv64-linux-gnu`, and the matching `riscv64gc -> riscv64` mapping in `scripts/build-npm.js`. Let me know if you'd rather handle it a different way.

gounthar:
Adds `riscv64gc-unknown-linux-gnu` to the release matrix so a prebuilt `lightningcss-linux-riscv64-gnu` gets published, and lightningcss loads natively on riscv64 Linux instead of throwing at require time.

Closes #1267.

### Changes

- `.github/workflows/release.yml`: new matrix entry on the existing napi-rs `lts-debian` image. It reuses the existing `matrix.setup` step to install `gcc-riscv64-linux-gnu`, strips with `riscv64-linux-gnu-strip`, and sets `CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER` on both build steps.
- `scripts/build-npm.js`: register the `riscv64gc-unknown-linux-gnu` triple (glibc) and map `riscv64gc -> riscv64`.

### Verification

- Cross-compile in the napi-rs container produces `lightningcss.linux-riscv64-gnu.node` (ELF 64-bit RISC-V) in about 3 minutes.
- A native build on a riscv64 runner produces the same artifact.
- Smoke test on real riscv64 hardware: `require()` loads it and `transform({...})` minifies `.a{ color: #ffffff }` to `.a{color:#fff}`.

I went with cross-compile to match the other targets rather than adding a native riscv64 runner to the matrix. Let me know if you'd prefer it done differently.


devongovett:
I'm not sure this has wide enough usage to take on officially. We also have no way to test that it works. What hardware are you running this on? Can you just compile it from source?

gounthar:
Hardware is a Banana Pi BPI-F3: SpaceMiT K1, 8 cores, Ubuntu 24.04, Node 20. A ~100 USD SBC, nothing exotic.

On usage, that's fair, riscv64 npm traffic is small today. The one thing that makes it surface a bit more than the raw numbers suggest is that nobody installs lightningcss deliberately, it arrives through vite or tailwind v4. So a normal frontend project on one of these boards dies at require time on a missing optional dependency, and the error names lightningcss rather than the thing that pulled it in. You have the download numbers and I don't, so you're better placed than me to judge whether that's worth a matrix entry.

Compiling from source does work, I did it: 53 minutes wall clock on the F3. The build isn't really the painful part, it's that the resulting .node has to be dropped into node_modules by hand after every install, in a package the user never chose.

On testing, RISE runs a pool of native riscv64 GitHub Actions runners (label `ubuntu-24.04-riscv`). It's free, access goes through a GitHub App install, and they've been granting it to projects that ask. That's what I used to validate the native build, so there is a way to test it on real hardware if you want one. I'm happy to add a smoke test job to this PR, or send it separately if you'd rather look at it on its own.

If the runner setup is the awkward part, I can introduce you to the RISE people who operate the pool and do the back and forth myself, so it stays off your plate. Their whole point is making riscv64 CI available to upstream projects.

And if you'd rather not carry the target at all, no hard feelings, I'll keep it downstream.

PR #1271, opened 2026-06-08
perf: add conditional rule CodSpeed benchmark

hardfist:
## Summary

Adds a CodSpeed benchmark fixture for long runs of adjacent conditional rules.

This introduces `stylesheet/transform/conditional-run`, which generates repeated adjacent `@media`, `@supports`, and `@container` blocks. The benchmark is intended to establish a baseline before applying the minification optimization in the follow-up PR.

## Validation

- `cargo check --benches`

### DIFF
--- a/benches/common/mod.rs
+++ b/benches/common/mod.rs
@@ -21,6 +21,10 @@ pub fn stylesheet_fixtures() -> Vec<Fixture> {
       name: "large",
       css: generated_fixture(192, 384),
     },
+    Fixture {
+      name: "conditional-run",
+      css: conditional_run_fixture(256),
+    },
   ]
 }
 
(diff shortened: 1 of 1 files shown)


kdy1 (approved):

kdy1:
I think this PR is good to go as it shows the performance difference, although Codspeed UI does not support this case.

ref: https://github.com/parcel-bundler/lightningcss/pull/1263

276.2 ms vs 13.9 ms


image

hardfist:
codspeed support manually compare two branches and it's 20x speed up for this case  https://app.codspeed.io/parcel-bundler/lightningcss/runs/compare/6a2688251fe9329263c497ff..6a23875a8d3fc83aa8269e47?uri=benches%2Ftransform.rs%3A%3Abenches%3A%3Abench_stylesheet_transform%3A%3Astylesheet%2Ftransform%255Bconditional-run%255D&runnerMode=Simulation
image

devongovett (approved):

PR #1274, opened 2026-06-09
feat: add `match-element` as an explicit ViewTransitionName variant

### LINKED ISSUE #1256: `view-transition-name: match-element` is incorrectly transformed
When using CSS Modules, LightningCSS transforms the standard CSS keyword `match-element` when used with `view-transition-name`.

Input:

```css
.example {
  view-transition-name: match-element;
}
```

Current output:

```css
.example {
  view-transition-name: SkZWNq_match-element;
}
```

Expected output:

```css
.example {
  view-transition-name: match-element;
}
```

It looks like the CSS Modules scoping/transformation logic currently treats `match-element` as a custom identifier, even though it is a reserved keyword defined by the spec: https://drafts.csswg.org/css-view-transitions/#view-transition-name-prop

PranavAchar01:
view-transition-name: none | <custom-ident> | match-element
https://drafts.csswg.org/css-view-transitions-2/#propdef-view-transition-name

Fixes: #1256


devongovett (approved):

PranavAchar01:
The failing `test-wasm` job looks unrelated to this PR. `master` fails the same job with an identical `rust-lld` link error (`undefined symbol: napi_get_cb_info` / `napi_typeof`), for example run 29718554844 on 2026-07-20. Happy to rebase if that would help pick up a fix.

Issue #1269, opened 2026-06-05
Error: [lightningcss minify] Cannot find module '../lightningcss.linux-arm64-musl.node'


peppergrayxyz:
Creating an empty Vite project and running it on arm64-musl fails with (https://github.com/vitejs/vite/issues/22494):

```bash
$ npm run build

> vite-project@0.0.0 build
> vue-tsc -b && vite build

vite v8.0.14 building client environment for production...
✓ 18 modules transformed.
✗ Build failed in 201ms
error during build:
Build failed with 1 error:

[plugin vite:css-post]
Error: [lightningcss minify] Cannot find module '../lightningcss.linux-arm64-musl.node'
Require stack:
- /home/pepper/dev/misc/vite-project/node_modules/lightningcss/node/index.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1456:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1066:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1071:22)
    at Module._load (node:internal/modules/cjs/loader:1242:25)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1556:12)
    at require (node:internal/modules/helpers:152:16)
    at Object.<anonymous> (/home/pepper/dev/misc/vite-project/node_modules/lightningcss/node/index.js:20:12)
    at Module._compile (node:internal/modules/cjs/loader:1812:14)
    at Object..js (node:internal/modules/cjs/loader:1943:10)
    at aggregateBindingErrorsIntoJsError (file:///home/pepper/dev/misc/vite-project/node_modules/rolldown/dist/shared/error-B8po7KiL.mjs:48:18)
    at unwrapBindingResult (file:///home/pepper/dev/misc/vite-project/node_modules/rolldown/dist/shared/error-B8po7KiL.mjs:18:128)
    at #build (file:///home/pepper/dev/misc/vite-project/node_modules/rolldown/dist/shared/rolldown-build-9MccaWPU.mjs:3236:34)
    at async buildEnvironment (file:///home/pepper/dev/misc/vite-project/node_modules/vite/dist/node/chunks/node.js:33137:64)
    at async Object.build (file:///home/pepper/dev/misc/vite-project/node_modules/vite/dist/node/chunks/node.js:33559:19)
[...]

PR #1282, opened 2026-06-18
fix: improve missing native binding error message for arm64 musl

renaudgenard:
## Summary
- report the expected native package and local fallback when Lightning CSS cannot load a platform binding
- include the original package and fallback load errors so arm64 musl installs point at `lightningcss-linux-arm64-musl` instead of only `../lightningcss.linux-arm64-musl.node`
- add a loader test that simulates a missing Linux arm64 musl optional dependency

Refs parcel-bundler/lightningcss#1269.

## Test plan
- `node --check node/index.js`
- `node --check node/test/native-binding.test.mjs`
- `npm exec -- uvu node/test native-binding.test.mjs`

### DIFF
--- a/node/test/native-binding.test.mjs
+++ b/node/test/native-binding.test.mjs
@@ -1,0 +1,47 @@
+import {test} from 'uvu';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import {fileURLToPath} from 'node:url';
+
+const dirname = path.dirname(fileURLToPath(import.meta.url));
+const indexPath = path.join(dirname, '..', 'index.js');
+const indexSource = fs.readFileSync(indexPath, 'utf8');
+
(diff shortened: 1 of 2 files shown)


kdy1 (commented):
I think this is a change that's pretty useful for edge cases, but the PR title is misleading.

node/test/native-binding.test.mjs
```diff
@@ -0,0 +1,47 @@
...
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import {fileURLToPath} from 'node:url';
+
+const dirname = path.dirname(fileURLToPath(import.meta.url));
+const indexPath = path.join(dirname, '..', 'index.js');
+const indexSource = fs.readFileSync(indexPath, 'utf8');
+
+function loadWithMissingNativeBinding(platform, arch, family) {
```
kdy1, line 12: Also, I really don't like this test

renaudgenard:
Thanks @kdy1, renamed the PR for clarity.

PR #1287, opened 2026-06-24
fix: preserve light-dark in custom properties

### LINKED ISSUE #821: `light-dark` is broken when children have different `color-scheme`'s
If you take the following HTML
```html
<p>This text should have a black background</p>
```

And apply the following CSS:
```css
:root {
  --background: light-dark(white, black);
  --text: light-dark(black, white);
}

p {
  color: var(--text);
  background: var(--background);
  color-scheme: dark;
}
```

The text changes to white text on black background, as expected because `color-scheme` is set to `dark`.

However, change this to the transpiled CSS generated by lightningcss
```css
:root{--background:var(--lightningcss-light,#fff)var(--lightningcss-dark,#000);--text:var(--lightningcss-light,#000)var(--lightningcss-dark,#fff)}p{color:var(--text);background:var(--background);--lightningcss-light: ;--lightningcss-dark:initial;color-scheme:dark}
```
And the text is black on white. I believe this is because the polyfill uses css variables which do not recompute if child elements change any of the input variables, where as `light-dark` seems to do. You currently need to re-apply the `:root` theme variables on every element that changes `color-scheme`

TheAlexLichter:
Resolves https://github.com/parcel-bundler/lightningcss/issues/821

This PR keeps the existing `light-dark()` polyfill for custom properties so old browser behavior and common design-system token patterns remain intact. 

But it also adds native `light-dark()` `@supports` overrides for supporting browsers which includes correct handling for LAB/P3 fallbacks, `!important`, `support` checks like custom properties or var()/attr() fallbacks and redeclarations.

`@property initial-value` preserves native `light-dark()` as is because otherwise it would invalidate typed custom-property registration.

Main downside is more generated CSS for custom properties with `light-dark()` but only when targeting browsers that need the polyfill behavior.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -182,6 +182,21 @@ impl<'i, 'o> PropertyHandlerContext<'i, 'o> {
     }
   }
 
+  pub fn remove_conditional_property(&mut self, property_id: &crate::properties::PropertyId) {
+    for entry in &mut self.supports {
+      let declarations = if self.is_important {
+        &mut entry.important_declarations
+      } else {
+        &mut entry.declarations
+      };
+      declarations.retain(|property| property.property_id() != *property_id);
+    }
+
+    self
+      .supports
+      .retain(|entry| !entry.declarations.is_empty() || !entry.important_declarations.is_empty());
+  }
+
   pub fn add_unparsed_fallbacks(&mut self, unparsed: &mut UnparsedProperty<'i>) {
     if self.context != DeclarationContext::StyleRule && self.context != DeclarationContext::StyleAttribute {
       return;
--- a/src/declaration.rs
+++ b/src/declaration.rs
@@ -626,6 +626,7 @@ impl<'i> DeclarationHandler<'i> {
           if self.decls[*index] == *property {
             return true;
           }
+          context.remove_conditional_property(&PropertyId::Custom(custom.name.clone()));
           let mut custom = custom.clone();
           self.add_conditional_fallbacks(&mut custom, context);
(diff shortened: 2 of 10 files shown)


devongovett:
I would say that's a breaking change. I know the polyfill isn't perfect, but people are relying on it anyway.

TheAlexLichter:
Back to the drawing board then. I'll try to come up with a solution that fixes light-dark() "better" while remaining backward-compatible with the current polyfill.

MoritzLost:
@devongovett @TheAlexLichter Is it a breaking change to fix something that's already broken, though?

If people are relying on this transpiler feature, they most likely are not using `color-scheme` in the way described in the issue, or they would've run into this bug. If they _did_ encounter the bug, they would have to have either disabled the feature, or implemented a workaround. In either case, nothing will break when the feature is only turned off in cases where it _would_ trigger the bug at the moment.

devongovett:
No, it's only broken in a very specific case when you set the color-scheme on a different element than the var. If you define your variables on the same element you set color-scheme it will work fine. This is a very common case, i.e. a design system usually sets it's tokens on a root-level element that also sets the color scheme. Removing this feature would break that.

Note that there is already a feature flag to disable LightDark if you want. We could consider a separate flag to disable it on vars I suppose.

TheAlexLichter:
Tried something more sophisticated (which is slightly ugly and gives`TokenList` a bit more to do 🙈).

Also fine with switching to a new feature flag instead if you think that approach isnt the way.

PR #1289, opened 2026-06-26
feat: support custom source map implemetation

intellild:
## Summary

This refactors CSS printing so `ToCss` depends on a sealed printer trait instead of the concrete `Printer<W, S>` type. It also splits source map writing behind `custom_sourcemap`, while keeping the existing `sourcemap` feature backed by Parcel source maps for current users.

## Why

Bundlers such as Rspack already have their own source map representation. Previously, they had to generate a Parcel source map from Lightning CSS output and then translate it back into their own source map type.

With this change, those consumers can provide a custom source map writer directly.

This has two benefits:

- Reduces binary size by avoiding the built-in Parcel source map implementation when it is not needed.
- Improves CSS printing performance by avoid source map translation.

In Rspack CI, this reduced the binary by about 400 KB:
https://github.com/web-infra-dev/rspack/actions/runs/28227945651/job/83627257783?pr=14581

Release builds are estimated to save around 300 KB.

## Compatibility

The default `sourcemap` feature still uses Parcel source maps, so existing Lightning CSS users should keep the same public behavior.

Consumers that want to avoid the built-in Parcel implementation can enable `custom_sourcemap` and provide their own writer.

## Validation

Checked the library with and without custom source maps, including no-default-feature and serde-only configurations.


devongovett:
Sounds like a good idea. I'll look more in depth soon.

intellild:
@devongovett Should we maintain rust API compatibility across the 1.0.0 beta releases?

devongovett:
no don't worry about that.

intellild:
@codex review

PR #1294, opened 2026-07-14
Add s390x-unknown-linux-gnu build target

satyamg1620:
Add s390x-unknown-linux-gnu build target


satyamg1620:
@devongovett @yisibl Can you please review this PR

devongovett:
I'm not sure this has wide enough usage to take on officially. We also have no way to test that it works. Can you just compile it from source on your hardware?

satyamg1620:
> I'm not sure this has wide enough usage to take on officially. We also have no way to test that it works. Can you just compile it from source on your hardware?

@devongovett I have compiled it on s390x and attached it
[s390x-compile-log.txt](https://github.com/user-attachments/files/30177822/s390x-compile-log.txt)

Let me know, if you want any additional things

PR #1299, opened 2026-07-30
Supports `:host-context` pseudo-class

ArtBoguslavskiy:
## Summary

Adds support for the `:host-context()` pseudo-class when parsing CSS selectors.

## Motivation

Ionic uses `:host-context()` in its component styles. When Lightning CSS is used as the CSS transformer in Vite, parsing fails with:

```text
[vite:css][lightningcss] 'host-context' is not recognized as a valid pseudo-class. Did you mean '::host-context' (pseudo-element) or is this a typo?
```

This prevents Lightning CSS from processing Ionic's generated CSS.

**P.S.** I'm not a Rust developer I'm primarily a front-end developer. This is my first contribution to a Rust project, so I hope my implementation is useful. If there's anything you'd like me to improve or implement differently, I'd be happy to update the PR.

### DIFF
--- a/selectors/builder.rs
+++ b/selectors/builder.rs
@@ -314,6 +314,11 @@ where
           *specificity += Specificity::from(selector.specificity());
         }
       }
+      Component::HostContext(ref selector) => {
+        specificity.class_like_selectors += 1;
+        // See: https://github.com/w3c/csswg-drafts/issues/1915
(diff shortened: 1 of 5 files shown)


devongovett:
looks deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:host-context

PR #1308, opened 2026-08-12
Fix parsing of pseudo-elements and pseudo-classes after ::details-content

### LINKED ISSUE #1244: Selecting `::details-content::before` pseudo element errors
I have some CSS that targets a `::before` pseudo element of the `<details>` element's `::details-content` using a selector containing `::details-content::before`. It seems to be supported in the major browsers (current versions of Firefox, Safari & Chrome parse and apply those styles - see screenshots below) so I presume this is valid CSS, however `lightningcss` errors with "Invalid state" when it encounters that.

You can reproduce this in the [Lightning CSS Playground](https://lightningcss.dev/playground/) with the following code:

```css
::details-content::before {
  background-color: red;
}
```

Image

Given that browsers accept this selector, I think Lightning CSS should be able to handle it and not error, so I think this is a bug.

---

As promised above, here a screenshot of the various browser dev tools showing this selector working fine:

**Firefox** (v150)
Image

**Chrome** (v148)
Image

**Safari** (v26.4)
Image

rorshopping:
Fixes #1244

## Problem

Selectors chaining pseudo-elements off `::details-content` fail to parse:

```css
::details-content::before { color: red }   /* error: Invalid state */
details::details-content::after { content: '' }
```

## Root cause

`::details-content` is an [element-backed pseudo-element](https://drafts.csswg.org/css-pseudo-4/#element-like): it behaves like a type selector, so other pseudo-elements and pseudo-classes may follow it. The parser did not model this distinction, and treated every known pseudo-element the same way as `::before` (which cannot be followed by further pseudo-elements).

## Changes

The relationship between pseudo-elements is now modeled comprehensively, mirroring the current upstream [`selectors`](https://github.com/servo/stylo/blob/main/selectors/parser.rs) parser in servo/stylo (where Firefox implements `::details-content`) and the Chromium change that allowed almost all pseudo-elements after `::part()` (crrev.com/c/5823774):

- Rename the parsing state to `AFTER_NON_ELEMENT_BACKED_PSEUDO`, add `AFTER_BEFORE_OR_AFTER_PSEUDO`, and add `PseudoElement::{parses_as_element_backed, is_before_or_after, valid_after_before_or_after}` trait methods, matching upstream naming.
- Element-backed pseudo-elements now parse exactly like `::part()`: all pseudo-classes (including functional ones like `:lang()`) and all other pseudo-elements may follow them, and they may also follow `::part()` itself, e.g. `::part(foo)::details-content::before`. `::part()`, `::slotted()`, combinators and type/class/id selectors remain errors after them.
- `::before`/`::after` accept a single following pseudo-element, restricted to a `valid_after_before_or_after()` allowlist that currently only contains `::marker` (per [css-pseudo-4](https://drafts.csswg.org/css-pseudo-4/#marker-pseudo)), so `::before::marker` and `::details-content::before::marker` parse while e.g. `::before::before`, `::marker::marker` and `::before::details-content` are still errors.
- lightningcss also marks `::picker()` as element-backed, matching Gecko (stylo's `pseudo_elements.toml`). The view-transition pseudo-elements are deliberately left out: css-view-transitions only allows `:only-child` after them.
- A bare element-backed pseudo-element now also sets the `HAS_PSEUDO` selector flag, so such selectors are excluded from the `:is()` lowering in `rules/mod.rs` (where pseudo-elements are not allowed).

Now all of these parse:

```css
::details-content::before { color: red }
::details-content::after { content: '' }
::details-content::marker { color: red }
::details-content::before::marker { color: red }
details::details-content::before { color: red }
details::details-content:lang(en)::before { content: '' }
::part(foo)::details-content::before { color: red }
::part(foo)::marker { color: red }
::picker(select)::before { color: red }
```

while these remain errors (matching browsers):

```css
::details-content div { color: red }          /* error */
::details-content.foo { color: red }          /* error */
::details-content::part(x) { color: red }     /* error */
::details-content::slotted(div) { color: red } /* error */
::details-content::before::after { color: red } /* error */
::before::before { color: red }               /* error */
::marker::marker { color: red }               /* error */
::view-transition-group(foo)::before { color: red } /* error */
```

## Tests

- selectors: `test_pseudo_on_element_backed_pseudo`, plus new `test_pseudo_duplicate_before_after_or_marker`, `test_pseudo_after_before_or_after` and `test_pseudo_after_part` (mirroring upstream), and `test_parsing` now covers the chains above.
- lightningcss: minify tests for `::details-content::before/after/marker`, `::before::marker`, `::details-content::before::marker`, `details::details-content::before`, `::part(foo)::details-content::before` and `::picker(select)::before`, plus error tests for `::details-content::part(foo)`, `::details-content::slotted(div)`, `::details-content::before::before`, `::details-content::before::selection`, `::before::details-content`, `::marker::marker` and `::part(foo)::part(bar)`.
- Also fixes the pre-existing `test_parsing` failures in the selectors crate (the `DummyParser` was missing `::target-text`, `::search-text`, `:current`, `::highlight()`, `::picker()`, `::picker-icon`, `::checkmark`, `::grammar-error` and `::spelling-error`), so that suite is fully green.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6822,6 +6822,19 @@ mod tests {
       "custom-element::part(foo) {color:red}",
       "custom-element::part(foo){color:red}",
     );
+    // https://github.com/parcel-bundler/lightningcss/issues/1244
+    minify_test("::details-content::before {color:red}", "::details-content:before{color:red}");
+    minify_test("::details-content::after {color:red}", "::details-content:after{color:red}");
+    minify_test("::details-content::marker {color:red}", "::details-content::marker{color:red}");
+    minify_test(
+      "details::details-content::before {color:red}",
+      "details::details-content:before{color:red}",
+    );
+    // https://github.com/parcel-bundler/lightningcss/issues/1244
+    error_test(
+      "::details-content::part(foo) {color:red}",
+      ParserError::SelectorError(SelectorError::InvalidState),
+    );
     minify_test(".sm\\:text-5xl { font-size: 3rem }", ".sm\\:text-5xl{font-size:3rem}");
     minify_test("a:has(> img) {color:red}", "a:has(>img){color:red}");
     minify_test("dt:has(+ dt) {color:red}", "dt:has(+dt){color:red}");
--- a/selectors/parser.rs
+++ b/selectors/parser.rs
@@ -48,6 +48,16 @@ pub trait PseudoElement<'i>: Sized + ToCss {
   fn is_unknown(&self) -> bool {
     false
   }
+
+  /// Whether this pseudo-element is element-backed.
+  ///
+  /// Element-backed pseudo-elements (such as `::details-content`) behave like
+  /// a type selector: pseudo-classes and other pseudo-elements may follow them.
+  ///
+  /// https://drafts.csswg.org/css-pseudo-4/#element-like
+  fn is_element_backed(&self) -> bool {
+    false
+  }
 }
 
 /// A trait that represents a pseudo-class.
@@ -122,8 +132,15 @@ bitflags! {
         /// as well.
         const AFTER_NON_STATEFUL_PSEUDO_ELEMENT = 1 << 4;
 
+        /// Whether we've parsed an element-backed pseudo-element (as in,
+        /// `::details-content`), which accepts tree-abiding pseudo-elements
+        /// and pseudo-classes after it.
+        ///
+        /// https://drafts.csswg.org/css-pseudo-4/#element-like
+        const AFTER_ELEMENT_BACKED_PSEUDO_ELEMENT = 1 << 11;
+
         /// Whether we are after any of the pseudo-like things.
-        const AFTER_PSEUDO = Self::AFTER_PART.bits() | Self::AFTER_SLOTTED.bits() | Self::AFTER_PSEUDO_ELEMENT.bits();
+        const AFTER_PSEUDO = Self::AFTER_PART.bits() | Self::AFTER_SLOTTED.bits() | Self::AFTER_PSEUDO_ELEMENT.bits() | Self::AFTER_ELEMENT_BACKED_PSEUDO_ELEMENT.bits();
 
         /// Whether we explicitly disallow combinators.
         const DISALLOW_COMBINATORS = 1 << 5;
@@ -163,7 +180,11 @@ impl SelectorParsingState {
   // state, and so on.
   #[inline]
   fn allows_custom_functional_pseudo_classes(self) -> bool {
-    !self.intersects(Self::AFTER_PSEUDO)
+    // Functional pseudo-classes are allowed after ::part and element-backed
+    // pseudo-elements, but not after ::slotted() or non-element-backed
+    // pseudo-elements.
+    // https://drafts.csswg.org/css-pseudo-4/#element-like
+    !self.intersects(Self::AFTER_SLOTTED | Self::AFTER_PSEUDO_ELEMENT)
   }
 
   #[inline]
@@ -2776,7 +2797,10 @@ where
         builder.push_simple_selector(Component::Slotted(selector));
       }
       SimpleSelectorParseResult::PseudoElement(p) => {
-        if !p.is_unknown() {
+        if p.is_element_backed() {
+          state.insert(SelectorParsingState::AFTER_ELEMENT_BACKED_PSEUDO_ELEMENT);
+          builder.push_combinator(Combinator::PseudoElement);
+        } else if !p.is_unknown() {
           state.insert(SelectorParsingState::AFTER_PSEUDO_ELEMENT);
           builder.push_combinator(Combinator::PseudoElement);
         } else {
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -6822,6 +6822,19 @@ mod tests {
       "custom-element::part(foo) {color:red}",
       "custom-element::part(foo){color:red}",
     );
+    // https://github.com/parcel-bundler/lightningcss/issues/1244
+    minify_test("::details-content::before {color:red}", "::details-content:before{color:red}");
+    minify_test("::details-content::after {color:red}", "::details-content:after{color:red}");
+    minify_test("::details-content::marker {color:red}", "::details-content::marker{color:red}");
```
yisibl, line 6828: What about `::details-content::before::marker` and `::part(foo)::details-content::before`?
See https://chromium-review.googlesource.com/c/chromium/src/+/5823774

yisibl: Overall, this requires a comprehensive consideration of the relationship with other pseudo-elements.

rorshopping: Good call — I've reworked this to comprehensively model the relationships between pseudo-elements, mirroring the current upstream [`selectors` parser in servo/stylo](https://github.com/servo/stylo/blob/main/selectors/parser.rs) (where Firefox implements `::details-content`, and where the `:lang()`-after-`::part()` fix landed) together with the rules from the Chromium CL you linked.

Both cases now parse:

- **`::details-content::before::marker`** — `::before`/`::after` now accept a single following pseudo-element, restricted to a new `valid_after_before_or_after()` allowlist that currently only contains `::marker` ([css-pseudo-4](https://drafts.csswg.org/css-pseudo-4/#marker-pseudo), and upstream stylo does the same). `::before::marker`, `::after::marker` and `::details-content::before::marker` parse; `::before::before`, `::marker::marker`, `::before::details-content` etc. remain errors.
- **`::part(foo)::details-content::before`** — element-backed pseudo-elements are now parsed exactly like `::part()` (upstream folds both into one state): they may follow `::part()` itself, and accept any pseudo-element/pseudo-class after them, matching the WPT tests from the CL (`::part(mypart)::details-content`, `::part(mypart)::marker`, …).

The resulting matrix:

| after | allowed | rejected |
| --- | --- | --- |
| `::part()`, element-backed (`::details-content`, `::picker()`) | pseudo-classes incl. functional (`:lang()`, `:dir()`), any pseudo-element incl. element-backed ones | `::part()`, `::slotted()`, combinators, type/class/id/attribute selectors |
| `::before`, `::after` | stateful pseudo-classes (`:hover`…), `::marker` only | any other pseudo-element, functional pseudo-classes |
| other pseudo-elements (e.g. `::marker`, `::selection`) | stateful pseudo-classes | any pseudo-element |

Other fixes that fell out of this:

- lightningcss now also marks `::picker()` as element-backed, matching Gecko (`pseudo_elements.toml`). The view-transition pseudo-elements are deliberately excluded: css-view-transitions only allows `:only-child` after them (Gecko's toml does parse them as element-backed via a default, but that conflicts with the spec's `:only-child`-only rule, so I kept them non-element-backed).
- A bare element-backed pseudo-element now sets the `HAS_PSEUDO` flag, so e.g. `::details-content` is no longer eligible for the `:is()` lowering in `rules/mod.rs` (`:is(::details-content)` would be invalid).
- Fixed the pre-existing `test_parsing` failures in the selectors crate (`DummyParser` was missing `::target-text`, `::search-text`, `::highlight()`, `::picker()` etc.), so that suite is fully green now.

One deliberate deviation from upstream stylo: after `::marker`, the `AFTER_BEFORE_OR_AFTER_PSEUDO` state is cleared again, so `::before::marker::marker` is rejected — upstream keeps the state set and would accept it, which looks like an oversight given their own `test_pseudo_duplicate_before_after_or_marker`.

rorshopping:
The failing `test-wasm` check is unrelated to the changes in this PR — it fails on other recent branches as well (e.g. @mischnic's `browserslist` branch, Sept 18), and I could reproduce it from this branch on a clean checkout with rustc 1.96.0 and 1.98.1 alike.

**Root cause:** rustc [1.96.0 stopped passing `--allow-undefined` to wasm-ld](https://releases.rs/docs/1.96.0) (see the [Rust blog post on WebAssembly changes](https://blog.rust-lang.org/2026/04/09/check-cfg-links-and-undefined-symbols/)). `napi-sys` declares the N-API functions as undefined `extern "C"` symbols that the host (Node) provides at runtime, so `yarn wasm:build-release` now fails linking `lightningcss_node` with `rust-lld: error: undefined symbol: napi_get_cb_info` etc. CI resolves `dtolnay/rust-toolchain@stable` to the latest stable (1.98.1), which overrides the `rust-toolchain.toml` pin via `RUSTUP_TOOLCHAIN`, so the job started failing even though nothing in the repo changed.

**Fix (0e2d60d):** pass `-C link-arg=--allow-undefined` explicitly for `wasm32-unknown-unknown` in the existing `.cargo/config.toml` `rustflags`, restoring exactly the flag rustc used to pass. Verified locally on both 1.96.0 and 1.98.1: the module links, exports `napi_register_module_v1`/`register_module`, and its 46 `napi_*` imports are again resolved by the host at runtime. This is also needed on master, since the same failure affects every branch built with a current toolchain.

PR #1316, opened 2026-08-26
fix(supports): keep the vendor prefix on a feature query inside and/or

### LINKED ISSUE #710: Improperly unprefixing -webkit-box-orient when in a @supports block
v1.24.1 is now removing the `-webkit-` prefix from `-webkit-box-orient`. Compare the output from [v1.24.0](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22%40supports%20(%5Cn%20%20%20%20(display%3A%20-webkit-box)%20and%20(-webkit-box-orient%3A%20vertical)%20and%20(-webkit-line-clamp%3A%203)%5Cn%20%20)%20%7B%5Cn%20%20.foo%20%7B%5Cn%20%20%20%20display%3A%20-webkit-box%3B%5Cn%20%20-webkit-box-orient%3A%20vertical%3B%5Cn%20%20-webkit-line-clamp%3A%203%3B%5Cn%20%20%7D%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%221.24.0%22%7D) to [1.24.1](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22%40supports%20(%5Cn%20%20%20%20(display%3A%20-webkit-box)%20and%20(-webkit-box-orient%3A%20vertical)%20and%20(-webkit-line-clamp%3A%203)%5Cn%20%20)%20%7B%5Cn%20%20.foo%20%7B%5Cn%20%20%20%20display%3A%20-webkit-box%3B%5Cn%20%20-webkit-box-orient%3A%20vertical%3B%5Cn%20%20-webkit-line-clamp%3A%203%3B%5Cn%20%20%7D%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%221.24.1%22%7D).

## Input
```css
@supports (
    (display: -webkit-box) and (-webkit-box-orient: vertical) and (-webkit-line-clamp: 3)
  ) {
  .foo {
    display: -webkit-box;
  -webkit-box-orient: vertical;
  -webkit-line-clamp: 3;
  }
}
```

## 1.24.0
```css
@supports (display: -webkit-box) and ((-webkit-box-orient: vertical)) and (-webkit-line-clamp: 3) {
  .foo {
    -webkit-line-clamp: 3;
    -webkit-box-orient: vertical;
    display: -webkit-box;
  }
}
```

## 1.24.1
```css
@supports (display: -webkit-box) and (box-orient: vertical) and (-webkit-line-clamp: 3) {
  .foo {
    -webkit-line-clamp: 3;
    -webkit-box-orient: vertical;
    display: -webkit-box;
  }
}
```

oddharsh:
Fixes #710.

The `@supports` merge pass binds the unprefixed id **over** `property_id`, and then uses that binding for two more things: what it stores, and which prefix it merges in.

```rust
let property_id = property_id.with_prefix(VendorPrefix::None);   // shadows
let key = (property_id.clone(), value.clone());
if let Some(index) = seen_declarations.get(&key) {
    cur_property.add_prefix(property_id.prefix());               // always VendorPrefix::None
} else {
    seen_declarations.insert(key, conditions.len());
    conditions.push(SupportsCondition::Declaration { property_id, value });   // stored stripped
}
```

So a condition matching nothing already seen is pushed with its prefix gone, and `(-webkit-box-orient: vertical)` becomes `(box-orient: vertical)` — which [no browser implements](https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient#browser_compatibility), so the block it guards silently never applies.

Using a separate binding for the match key fixes both symptoms. The unprefixed id is the **key** only; the condition keeps the prefix it was written with, and the merge branch adds the real prefix instead of `None`.

## Why it reads as an inconsistency rather than a plain regression

A prefixed query survives when it stands **alone**, because that path pushes `in_parens` untouched. It is only stripped as an operand of `and` / `or`. Measured on 1.33.0 before this change:

| condition | alone | inside `(display:flex) or …` |
|---|---|---|
| `(-webkit-box-orient:vertical)` | preserved | `(box-orient:vertical)` |
| `(-webkit-backdrop-filter:blur(1px))` | preserved | `(backdrop-filter:blur(1px))` |
| `(-moz-appearance:none)` | preserved | `(appearance:none)` |

Worth knowing for anyone triaging this: reducing a report to a single condition to check whether it still reproduces answers **no** on a build that is still broken.

## Why the existing tests did not catch it

Every assertion added with the merge feature in 6bd2761 sets `Browsers` targets, and the prefixing pass then re-adds what the merge dropped. The regression tests here deliberately set none, and they use `minify_test` rather than `prefix_test` for that reason.

## Verification

Before and after, same input, `minify: true`, no targets:

```
BEFORE   @supports (display:flex) and (box-orient:vertical){a{color:red}}
AFTER    @supports (display:flex) and ((-webkit-box-orient:vertical)){a{color:red}}
```

- 120/120 lib tests pass, including the merge assertions from 6bd2761
- the three added tests **fail without the fix**, printing the stripped form, and pass with it
- `cargo fmt` leaves both changed files untouched (it does reformat `bundler.rs`, `selector.rs` and an unrelated `lib.rs` line under current rustfmt; those are deliberately not included here)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15078,6 +15078,28 @@ mod tests {
 
   #[test]
   fn test_supports_rule() {
+    // A vendor-prefixed feature query must keep its prefix when it is an OPERAND
+    // of `and` / `or`, not only when it stands alone. The merge pass bound the
+    // unprefixed id over `property_id` and then stored THAT, so a condition
+    // matching nothing already seen was pushed stripped: `-webkit-box-orient`
+    // became `box-orient`, which no browser implements, silently disabling the
+    // whole block. Regression test for #710.
+    //
+    // These carry no `Browsers` targets on purpose. Every existing assertion for
+    // this merge sets them, and the prefixing pass then re-adds what was lost,
+    // which is why the bug survived the tests that shipped with the feature.
+    minify_test(
+      "@supports (display: flex) and (-webkit-box-orient: vertical) { .a { color: red } }",
+      "@supports (display:flex) and ((-webkit-box-orient:vertical)){.a{color:red}}",
+    );
+    minify_test(
+      "@supports (display: flex) or (-moz-appearance: none) { .a { color: red } }",
+      "@supports (display:flex) or ((-moz-appearance:none)){.a{color:red}}",
+    );
+    minify_test(
+      "@supports (display: -webkit-box) and (-webkit-box-orient: vertical) and (-webkit-line-clamp: 3) { .a { color: red } }",
+      "@supports (display:-webkit-box) and ((-webkit-box-orient:vertical)) and (-webkit-line-clamp:3){.a{color:red}}",
+    );
     test(
       r#"
       @supports (foo: bar) {
--- a/src/rules/supports.rs
+++ b/src/rules/supports.rs
@@ -236,8 +236,12 @@ impl<'i> Parse<'i> for SupportsCondition<'i> {
 
         if let SupportsCondition::Declaration { property_id, value } = condition {
           // Merge multiple declarations with the same property id (minus prefix) and value together.
-          let property_id = property_id.with_prefix(VendorPrefix::None);
-          let key = (property_id.clone(), value.clone());
+          // The unprefixed id is the MATCH KEY only. Binding it over `property_id` also changed
+          // what got stored and what prefix got merged in, so a condition that matched nothing was
+          // pushed with its prefix stripped -- turning `(-webkit-box-orient: vertical)` into
+          // `(box-orient: vertical)`, which no browser implements.
+          let unprefixed = property_id.with_prefix(VendorPrefix::None);
+          let key = (unprefixed, value.clone());
           if let Some(index) = seen_declarations.get(&key) {
             if let SupportsCondition::Declaration {
               property_id: cur_property,


oddharsh:
Re-verified on master (c6a0c3c) today, and the trigger is narrower than the description here says, which may help whoever picks this up. Built the CLI from master, unpatched, `--minify`:

| source condition | output | prefix |
|---|---|---|
| `@supports (-moz-appearance: none)` | `@supports ((-moz-appearance:none))` | kept |
| `@supports (-moz-appearance: none) or (display: flex)` | `@supports ((-moz-appearance:none)) or (display:flex)` | kept |
| `@supports (not selector(::-webkit-scrollbar)) or (-moz-appearance: none)` | `@supports (not selector(::-webkit-scrollbar)) or (appearance:none)` | **stripped** |

So it is not "any compound condition": the prefixed condition survives in first position and is rewritten when it follows another operand, which matches the merge-pass cause in the PR body (the binding is made while merging a later condition into what came before).

That third row is the shape that bit us. `@supports (not selector(::-webkit-scrollbar)) or selector(:-moz-focusring)` is how a stylesheet reaches Firefox alone for the `scrollbar-color` / `::-webkit-scrollbar` interaction, and `-moz-appearance` is the obvious thing to reach for there. Rewritten to `appearance`, Chromium answers the query true and the Firefox-only arm applies everywhere, silently.

Happy to add that row as a test case, or to rebase if the shape of the fix needs changing.

PR #1329, opened 2026-09-05
Add NonNegative wrapper for numeric values

### LINKED ISSUE #1000: The conversion of calc() does not take into account the valid range of the property value
## Input

```css
.foo {
  width: calc(5px - 10px);
}
```

## Lightningcss output

```css
.foo {
  width: -5px;
}
```

## Expected

```css
.foo {
  width: 0px;
}
```

## Specs

https://drafts.csswg.org/css-values-4/#example-ec14dee2

> Note however that [width: -5px](https://drafts.csswg.org/css-sizing-3/#propdef-width) is not equivalent to width: calc(-5px)! Out-of-range values outside [calc()](https://drafts.csswg.org/css-values-4/#funcdef-calc) are syntactically invalid, and cause the entire declaration to be dropped.

It is very **unsafe** to do unfolding calculations for calc(), we should add an option to turn it off.

devongovett:
Fixes #1000. Also fixes over 600 web platform tests (run against #1328).

Properties that only allow non-negative values are now wrapped in the `NonNegative` helper type, e.g. `NonNegative<Length>`. This rejects invalid negative values during parsing (though we still fall back to preserving the raw css tokens). For known calc expressions that resolve to negative values e.g. `calc(-10px)`, these are still syntactically valid, but are now clamped to zero.

### DIFF
--- a/src/bundler.rs
+++ b/src/bundler.rs
@@ -106,10 +106,7 @@ pub enum ResolveResult {
   /// An external URL.
   External(String),
   /// A file path.
-  #[cfg_attr(
-    any(feature = "serde", feature = "nodejs"),
-    serde(untagged)
-  )]
+  #[cfg_attr(any(feature = "serde", feature = "nodejs"), serde(untagged))]
   File(PathBuf),
 }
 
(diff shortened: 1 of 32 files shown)


yisibl:
Nice, it looks fine to me.

PR #1332, opened 2026-09-07
Improving floating-point precision in CSS color functions

yisibl:
This PR improves the precision of floating-point calculations, allowing us to pass more WPT test cases.

At the same time, performance has also been optimized.

`hsl_to_rgb()` has also been updated to use the algorithm recommended in the CSS specification
 https://drafts.csswg.org/css-color-4/#hsl-to-rgb


src/lib.rs
```diff
@@ -18308,6 +18378,7 @@ mod tests {
     minify_test(".foo { color: hwb(194 50% 0%) }", ".foo{color:#80e1ff}");
     minify_test(".foo { color: hwb(194 50 0) }", ".foo{color:#80e1ff}");
     minify_test(".foo { color: hwb(194 50% 50%) }", ".foo{color:gray}");
+    minify_test(".foo { color: hwb(320deg 30% 40%) }", ".foo{color:#994d80}");
```
yisibl, line 18393: Before the fix, the output was:
```css
.foo{color:#994d7f}
```

PR #1333, opened 2026-09-08
fix: factor shared custom properties in adjacent rules

SyMind:
Adjacent theme rules often repeat many custom properties while differing in a few others. The existing rule merger keeps both copies unless the entire declaration blocks match. This change extracts matching custom properties into a shared selector rule when doing so reduces the serialized CSS size.

Extraction compares names, values, and importance, including the cost of additional selectors and separators. It operates on adjacent leaf rules from the same source, skips CSS Modules, actual vendor prefixes, and nesting, and leaves ordinary declarations in place. A local worklist revisits changed neighbors so newly adjacent shared rules and identical remaining blocks can merge. Normal and important overrides are resolved independently without regenerating property fallbacks.

Selector compatibility checks now inspect non-forgiving arguments of `:host()`, `::slotted()`, and `::cue()`; unsupported arguments must not invalidate a valid selector when combined into a list. Nesting detection also visits `:has()` and other function arguments that the base visitor does not traverse. Empty prefix flags and the unprefixed flag can merge, while actual vendor prefixes remain excluded.

The regression tests use formatted CSS and cover 44 scenarios plus repeat-pass assertions: declaration order, overrides, importance, variable-name case sensitivity, conditional boundaries, keyframes, selector compatibility, nesting, CSS Modules, byte-cost thresholds, and predecessor revisits.

On the motivating stylesheet, compared with `c6a0c3c` using identical explicit browser targets:

| Size | Before | After | Change |
| --- | ---: | ---: | ---: |
| CSS | 154,731 B | 135,993 B | −18,738 B (−12.11%) |
| gzip (level 9) | 45,674 B | 43,194 B | −2,480 B |
| Brotli (quality 11) | 34,308 B | 34,375 B | +67 B |

The heuristic optimizes raw CSS size; compressed size can vary, as the Brotli result shows.

Validation:

- Original theme regression and the selector-argument, nested-`:has()`, and predecessor-revisit regressions confirmed failing before their fixes.
- `RUSTFLAGS='-D warnings' cargo test --all-features --locked`: 169 tests passed, including integration and documentation tests.
- `yarn build`, `yarn test` (81 passed), and `yarn tsc` passed.
- Chromium 152: 276 fixtures, including 250 deterministic generated custom-property cases; 12,887 computed-style values compared with the base version, with no differences. No raw-size increases in this sample.
- The original stylesheet's output is unchanged by the additional boundary fixes; the earlier comparison of its 1,589 theme variables also passed.

Scope: this local worklist does not promise whole-compiler idempotence. Two of the 276 samples shrink when every compiler pass is run again: one also does so on the base version; in the other, extraction exposes a redundant non-adjacent rule that the existing global duplicate pass only recognizes on the next run. Updating that global pass and choosing a globally optimal extraction order are outside this change.


SyMind:
@codex review

PR #1341, opened 2026-09-08
fix: remove redundant initial padding after zero

SyMind:
When all four physical padding sides are already zero, a following `padding: initial` is redundant but currently survives minification. Keep the explicit zero declaration and omit that redundant reset:

```css
/* Input */
.foo {
  padding: 0;
  padding: initial;
}

/* Output */
.foo{padding:0}
```

This is limited to a parsed `initial` identifier after four known zero pixel lengths, with no pending logical padding state. It preserves the original flush boundary so later declarations, including legacy logical aliases, keep their order. Keeping `padding: 0` also preserves the fallback for browsers that do not support `initial`.

Validation:

- `RUSTFLAGS='-D warnings' cargo test --all-features --locked` passes.
- Regression coverage includes missing/nonzero sides, percentage zero, logical properties, variables, `all`, other CSS-wide keywords, mixed importance, other side shorthands, escaped/case-insensitive keywords, legacy aliases, formatted output, and IE 11/modern targets.
- `yarn build`, all 81 Node tests, and `yarn tsc` pass.
- Original and minified CSS have equal computed padding, margin, scroll padding, and inset values across 296 comparisons per engine in Chromium 151, Firefox 153, and WebKit 26.5, covering LTR/RTL and horizontal/vertical writing modes.


yisibl:
Thank you for your contribution.

I think that if we’re going to make changes like this, we need to consider a unified solution—it’s not just about padding; other properties have similar issues as well.

However, I think this is a bit too radical. As CSS becomes increasingly complex, it’s unclear whether there might be any side effects. This kind of optimization should be left up to the users to decide for themselves.

